PR #7, opened 2019-05-06
Use local images when docker daemon exists

knqyf263:
Use local images when docker daemon exists in the local machine.
If not, fanal will download image layers from a remote repository.

### DIFF
--- a/analyzer/analyzer.go
+++ b/analyzer/analyzer.go
@@ -92,9 +92,19 @@ func RequiredFilenames() []string {
 
 func Analyze(ctx context.Context, imageName string) (filesMap extractor.FileMap, err error) {
 	e := extractor.NewDockerExtractor(extractor.DockerOption{Timeout: 600 * time.Second})
-	filesMap, err = e.Extract(ctx, imageName, RequiredFilenames())
+	r, err := e.SaveLocalImage(ctx, imageName)
 	if err != nil {
-		return nil, errors.Wrap(err, "Failed to extract files")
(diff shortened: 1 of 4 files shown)


extractor/docker.go
```diff
@@ -127,6 +128,19 @@ func (d DockerExtractor) createRegistryClient(ctx context.Context, domain string
 	})
 }
 
+func (d DockerExtractor) SaveLocalImage(ctx context.Context, imageName string) (io.ReadCloser, error) {
+	cli, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return nil, xerrors.Errorf("error in docker NewClient")
```
tomoyamachi: It seems good enough to use xerrors.New.

analyzer/analyzer.go
```diff
@@ -92,9 +92,19 @@ func RequiredFilenames() []string {
...
+		filesMap, err = e.Extract(ctx, imageName, RequiredFilenames())
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to extract files")
+		}
+		return filesMap, nil
+	}
+
+	filesMap, err = e.ExtractFromFile(ctx, r, RequiredFilenames())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to extract files from saved tar")
```
tomoyamachi: It's better to use xerrors.Errorf() instead of errors.Wrap.

extractor/docker.go
knqyf263: Thanks!

Issue #28, opened 2019-09-02
Scan /var/lib/dpkg/status.d/xxx files for dpkg packages


tomoyamachi:
https://github.com/aquasecurity/trivy/issues/96
This issue caused by missing `/var/lib/dpkg/status` file.
Distroless images have some status files in `/var/lib/dpkg/status.d/`.

PR #29, opened 2019-09-02
extract all files in target require filedirs

tomoyamachi:
fix for https://github.com/aquasecurity/trivy/issues/96 (https://github.com/aquasecurity/fanal/issues/28)

```
$ go run cmd/fanal/main.go gcr.io/distroless/java:11
{Name:9.9 Family:debian}
via image Packages: 19
via file Packages: 0

$ go run cmd/fanal/main.go debian:latest
{Name:10.0 Family:debian}
via image Packages: 91
via file Packages: 0
```

### DIFF
--- a/analyzer/pkg/dpkg/dpkg.go
+++ b/analyzer/pkg/dpkg/dpkg.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"log"
-	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -32,7 +31,7 @@ func (a debianPkgAnalyzer) Analyze(fileMap extractor.FileMap) (pkgs []analyzer.P
 	detected := false
 	for _, filename := range a.RequiredFiles() {
 		// find by target directory
-		if filename[len(filename)-1:] == string(os.PathSeparator) {
+		if filename[len(filename)-1:] == "/" {
 			for targetname, targetbytes := range fileMap {
 				if filepath.Dir(filename) == filepath.Dir(targetname) {
 					scanner := bufio.NewScanner(bytes.NewBuffer(targetbytes))
--- a/analyzer/pkg/dpkg/dpkg.go
+++ b/analyzer/pkg/dpkg/dpkg.go
@@ -8,6 +8,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/aquasecurity/fanal/utils"
+
 	mapset "github.com/deckarep/golang-set"
 
 	"github.com/aquasecurity/fanal/analyzer"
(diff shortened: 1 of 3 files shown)


analyzer/pkg/dpkg/dpkg.go
```diff
@@ -29,6 +30,19 @@ type debianPkgAnalyzer struct{}
 func (a debianPkgAnalyzer) Analyze(fileMap extractor.FileMap) (pkgs []analyzer.Package, err error) {
 	detected := false
 	for _, filename := range a.RequiredFiles() {
+		// find by target directory
+		if filename[len(filename)-1:] == "/" {
```
knqyf263:
```suggestion
		if filename[len(filename)-1] == '/' {
```

extractor/docker/docker.go
```diff
@@ -336,6 +336,14 @@ func (d DockerExtractor) ExtractFiles(layer io.Reader, filenames []string) (extr
 		// Determine if we should extract the element
 		extract := false
 		for _, s := range filenames {
+			// extract all files in target directory if last char is "/"(Separator)
+			if s[len(s)-1:] == "/" {
```
knqyf263: Ditto

extractor/docker/docker.go
```diff
@@ -336,6 +336,14 @@ func (d DockerExtractor) ExtractFiles(layer io.Reader, filenames []string) (extr
...
+				if filepath.Dir(s) == filepath.Dir(filePath) {
```
knqyf263: Is filepath.Dir(s) necessary? Already checked.

```suggestion
				if s == filepath.Dir(filePath) {
```

analyzer/pkg/dpkg/dpkg.go
```diff
@@ -29,6 +30,19 @@ type debianPkgAnalyzer struct{}
...
+			for targetname, targetbytes := range fileMap {
+				if filepath.Dir(filename) == filepath.Dir(targetname) {
```
knqyf263: Ditto

analyzer/pkg/dpkg/dpkg.go
```diff
@@ -144,5 +158,5 @@ func (a debianPkgAnalyzer) parseDpkgPkg(scanner *bufio.Scanner) (pkg *analyzer.P
 }
 
 func (a debianPkgAnalyzer) RequiredFiles() []string {
-	return []string{"var/lib/dpkg/status"}
+	return []string{"var/lib/dpkg/status", "var/lib/dpkg/status.d/"}
```
knqyf263:
```suggestion
	return append(a.requiredFiles(), a.requiredDirs()...)

func (a debianPkgAnalyzer) requiredFiles() []string {
	return []string{"var/lib/dpkg/status"}
}	
func (a debianPkgAnalyzer) requiredDirs() []string {
	return []string{"var/lib/dpkg/status.d/"}
}
```

If you care about a performance, we can define these variables in advance.

analyzer/pkg/dpkg/dpkg.go
```diff
@@ -29,6 +30,19 @@ type debianPkgAnalyzer struct{}
 func (a debianPkgAnalyzer) Analyze(fileMap extractor.FileMap) (pkgs []analyzer.Package, err error) {
 	detected := false
 	for _, filename := range a.RequiredFiles() {
+		// find by target directory
+		if filename[len(filename)-1:] == "/" {
+			for targetname, targetbytes := range fileMap {
```
knqyf263: It seems better to make this loop the outer.

Before
```
for _, filename := range a.RequiredFiles()
    for targetname, targetbytes := range fileMap
```

After
```
for targetName, targetBytes := range fileMap
    dir := filepath.Dir(targetName)
    if !utils.StringInSlice(targetName, a.requiredFiles()) && !utils.StringInSlice(dir, a.requiredDirs()){
        continue
    }
```

extractor/docker/docker.go
tomoyamachi: s : `var/lib/dpkg/status.d/`
filepath.Dir(s) : `var/lib/dpkg/status.d`

tomoyamachi: `s[0:len(s)-2] == filepath.Dir(filePath)` looks mess.

knqyf263: I got it. However, we already know `s` is a directory. It seems a little weird to apply filepath.Dir() to `s`. We may need a comment.

knqyf263: How about `filepath.Clean`?

knqyf263: >Dir calls Clean on the path and trailing slashes are removed.
https://golang.org/pkg/path/filepath/#Dir

knqyf263: We must remove trailing slashes rather than extracting directory name. I wanna make it obvious.

tomoyamachi: I fixed it.
I can change OSAnalyzer interface to following if you want.

```
type OSAnalyzer interface {
	Analyze(extractor.FileMap) (OS, error)
	RequiredFiles() []string
	RequiredDirs() []string
}
```

analyzer/pkg/dpkg/dpkg.go
tomoyamachi: Currently, I added "/" to last char.

```
dir := filepath.Dir(filename) + "/"
```

If you don't like, I can create RequiredDir func.
check => https://github.com/aquasecurity/fanal/pull/29#discussion_r320302883

knqyf263:
Thank you for applying my comments. We need some unittests 😇 
Add image5.tar and directory to filenames:
https://github.com/aquasecurity/fanal/blob/master/extractor/docker/docker_test.go

PR #38, opened 2019-09-25
Fix wrong break

knqyf263:
Fix #29 
If `RequiredFiles()` contains a directory (which has a slash at the end), it always breaks the loop and can't extract files correctly.


knqyf263:
I agree with you. Thank you.

PR #39, opened 2019-09-25
Check whether rpm is installed

knqyf263:
For an easy-to-understand message
https://github.com/aquasecurity/trivy/issues/156

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -32,8 +32,7 @@ func StringInSlice(a string, list []string) bool {
 }
 
 func IsCommandAvailable(name string) bool {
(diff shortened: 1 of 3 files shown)


utils/utils.go
```diff
@@ -29,3 +30,11 @@ func StringInSlice(a string, list []string) bool {
 	}
 	return false
 }
+
+func IsCommandAvailable(name string) bool {
+	cmd := exec.Command(name, "--help")
```
danielpacak: NIT: I was wondering if exec.LookupPath("rpm") does the same job.

knqyf263: I fixed it. Thanks.

PR #40, opened 2019-10-14
feat(image): support tar.gz image

knqyf263:
This PR enables the automatic detection of file type (only gzip). If `xxx.tar` is passed to fanal, it is decompressed as a tar file. If `yyy.tar.gz` is passed, it is detected automatically and decompressed as a gzip file.


knqyf263:
@danielpacak Thanks!

PR #43, opened 2019-10-23
Detect Gentoo Linux

mrueg:
Currently detects the OS, planning to create an analyzer that detects installed packages from /var/db/pkg

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ import (
 	_ "github.com/aquasecurity/fanal/analyzer/os/amazonlinux"
 	_ "github.com/aquasecurity/fanal/analyzer/os/debianbase"
 	_ "github.com/aquasecurity/fanal/analyzer/os/suse"
+	_ "github.com/aquasecurity/fanal/analyzer/os/gentoo"
 	_ "github.com/aquasecurity/fanal/analyzer/os/redhatbase"
 	_ "github.com/aquasecurity/fanal/analyzer/pkg/apk"
 	_ "github.com/aquasecurity/fanal/analyzer/pkg/dpkg"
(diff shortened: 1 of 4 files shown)


mrueg:
This will detect Gentoo Linux, no package scanning available yet.

mrueg:
@knqyf263 let me know if you feel like this is worth merging. Otherwise I'll probably close the PR. :)

knqyf263:
I'm sorry for the late reply. I thought you would add an analyzer for OS packages. Is it difficult?

mrueg:
It turns out to be a bit more difficult to add it when I attempted it.
Information on installed packages are not stored in a single file, but in a directory structure in `/var/db/pkg/${CATEGORY}/${PACKAGENAME}-${PACKAGEVERSION}`, so in order to parse that info, one would need to go through all the directories and read the directory names.

Then you would have a list of packages like `net-vpn/openvpn-2.4.9`, which probably are difficult to map against any vuln db that trivy supports. 

So eventually there are two solutions for that:
a) Gentoo has a metadata.xml file in their "ebuild" tree, that contains more information. Challenge here is, this is usually stripped from container images due to the size (adds roughly 1GB to the image). If it were available that could be used to map packages to CVEs, as it contains an optional attribute:
```
 <upstream>
    <remote-id type="cpe">cpe:/a:openvpn:openvpn</remote-id>
</upstream>
```
b) Gentoo has it's own vulnerability management with Gentoo Linux Security Advisories (GLSA), this fetches xml files and uses a tool called glsa-check to compare against the content of /var/db/pkg/* 
trivy could include this in its vuln db and then compare the discovered packages against it.

PR #44, opened 2019-10-30
feat(library): ignore files under vendor dir

knqyf263:
fanal ignores `package-lock.json` under `node_modules` dir and composer.lock under `vendor` dir. But, composer.lock may rarely be under `node_modules`.
https://github.com/aquasecurity/trivy/issues/179#issuecomment-547413735

So, I think fanal should ignore all files under `node_modules` and `vendor` directory.


knqyf263:
Thanks.

PR #46, opened 2019-11-16
Add cache tests

simar7:
Adds some unit tests for cache.go

### DIFF
--- a/cache/cache_test.go
+++ b/cache/cache_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-func TestSetAndGet(t *testing.T) {
+func TestSetAndGetAndClear(t *testing.T) {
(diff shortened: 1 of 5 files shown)


cache/cache_test.go
```diff
@@ -0,0 +1,41 @@
...
+	assert.Equal(t, expectedCacheContents, string(b))
+
+
+	// get
+	actualFile := Get(f.Name())
+	actualBytes, _ := ioutil.ReadAll(actualFile)
+	assert.Equal(t, expectedCacheContents, string(actualBytes))
+
+	// clear
+	assert.NoError(t, Clear())
```
knqyf263: After `Clear()`, it may be better that we confirm the dir/file is removed.

simar7: Good point. I've added that.

PR #48, opened 2019-11-21
fix docker reg with option SkipPing

jabielecki:
Workaround for a deficient Ping implementation of reg package.
Ping fails on docker registries that return http 401
Authentication Required when requesting general /v2 url, but
happily allow unauthenticated pull of a specific image.

If there is any benefit of having Ping() enabled, it's not apparent.
Especially comparing it with how many registries are unusable with it.

Closes aquasecurity/trivy#229

Signed-off-by: Jakub Bielecki <jakub.bielecki@codilime.com>

### DIFF
--- a/analyzer/analyzer.go
+++ b/analyzer/analyzer.go
@@ -116,7 +116,8 @@ func RequiredFilenames() []string {
 func Analyze(ctx context.Context, imageName string, opts ...types.DockerOption) (fileMap extractor.FileMap, err error) {
 	// default docker option
 	opt := types.DockerOption{
-		Timeout: 600 * time.Second,
+		Timeout:  600 * time.Second,
+		SkipPing: true,
 	}
 	if len(opts) > 0 {
 		opt = opts[0]


jabielecki:
Before change:
```
$ go run ./cmd/fanal quay.io/prometheus/node-exporter
2019/11/21 22:03:58 failed to extract files: failed to create the registry client: Get https://quay.io/v2/: http: non-successful response (status=401 body="{\"error\": \"Invalid bearer token format\"}")
exit status 1
```

After this change:
```
$ go run ./cmd/fanal quay.io/prometheus/node-exporter
2019/11/21 22:12:17 Unknown OS
exit status 1
```

jabielecki:
Another benefit after the change is Azure CR:
```
$ go run ./cmd/fanal aksrepos.azurecr.io/prod/hcp-tunnel-front:v1.9.2-v3.0.7
{Name:18.04 Family:ubuntu}
via image Packages: 214
via file Packages: 0
```

knqyf263:
Thanks a lot!!

PR #50, opened 2019-11-21
change mod genuinetools/reg to vanilla

jabielecki:
Instead of using tomoyamachi's fork we can now use the vanilla upstream
package genuinetools/reg. This package gets better maintenance.

Also introducing new checksums for reg's children/dependecies.

Signed-off-by: Jakub Bielecki <jakub.bielecki@codilime.com>


knqyf263:
@jabielecki Thank you! Would you apply `go mod tidy`?

jabielecki:
With pleasure sir, `go mod tidy` there is. (Tested) @knqyf263

jabielecki:
I like this project. I'll be contributing more this and next month :)

knqyf263:
@jabielecki Thanks a lot!

PR #51, opened 2019-11-22
Cache: Save only required files

simar7:
This PR introduces the ability to store the database as a KV store instead of using the filesystem directly.

In addition it also adds:
* Refactoring
* Seams for testing
* Dependency injection for said seams
* More granular unit tests

I realize this PR is big, so I have split it up into two sections:
* Refactoring and adding seams: https://github.com/aquasecurity/fanal/pull/47
* Using KV store: https://github.com/aquasecurity/fanal/pull/51


knqyf263:
This commit contains the duplicated commits with #47. If this is a stacked PR, the target branch should be `adding-seams`.

simar7:
hey @knqyf263 I simplified a few things. After https://github.com/aquasecurity/fanal/pull/47 is merged, we can review and merge this one!

simar7:
hey @knqyf263 one caveat I just found with storing only required files is that something like this that currently works (even though ubuntu is not supported) won't work in the future as there are no required files being passed on as arguments:

```
➜  fanal git:(save-only-required-files) ✗ ./fanal -clear ubuntu #firstrun
file:  var/lib/dpkg/status
file:  etc/debian_version
file:  etc/lsb-release
file:  etc/os-release
file:  usr/lib/os-release
file:  /config
{Name:18.04 Family:ubuntu}
via image Packages: 89
via file Packages: 0
➜  fanal git:(save-only-required-files) ✗ ./fanal  ubuntu   #secondrun
file:  /config
2019/11/27 14:53:13 Unknown OS
```

What do you think about that? Since `ubuntu` is not currently supported, should we care about this?

knqyf263:
@simar7 What is happening in your example? And, what do you mean `ubuntu` is not supported?

extractor/docker/docker.go
```diff
@@ -145,21 +158,41 @@ func (d Extractor) createRegistryClient(ctx context.Context, domain string) (*re
...
+
+	var storedImage []byte
+	found, err := d.Cache.Get(kvtypes.GetItemInput{
+		BucketName: "imagebucket",
+		Key:        imageName,
+		Value:      &storedImage,
+	})
+
+	if found {
+		return ioutil.NopCloser(bytes.NewReader(storedImage)), nil
```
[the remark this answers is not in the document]
simar7: Correct, this is cruft.

simar7:
storing two images, amazonlinux + opensuse/leap in cache:
fanal-master:    100MB
fanal-bolt-gzip: 8MB
fanal-bolt-zstd: 4MB

extractor/docker/docker.go
```diff
@@ -145,21 +164,41 @@ func (d Extractor) createRegistryClient(ctx context.Context, domain string) (*re
...
+
+	var savedImage []byte
+	if err != nil || !found {
+		storedReader, err = d.saveLocalImage(ctx, imageName)
 		if err != nil {
 			return nil, xerrors.Errorf("failed to save the image: %w", err)
 		}
-		r, err = d.Cache.Set(imageName, r)
+
+		savedImage, err = ioutil.ReadAll(storedReader)
```
knqyf263: Some images are so large. I'd like to avoid copying an entire layer into memory. It may be better to extract only the necessary files from the image.

extractor/docker/docker.go
```diff
@@ -145,21 +164,41 @@ func (d Extractor) createRegistryClient(ctx context.Context, domain string) (*re
...
 		if err != nil {
-			log.Print(err)
+			return nil, xerrors.Errorf("failed to read saved image: %w", err)
+		}
+
+		if err := d.Cache.BatchSet(kvtypes.BatchSetItemInput{
```
knqyf263: We also need compression for an image.

extractor/docker/docker_test.go
```diff
@@ -187,35 +208,70 @@ func TestExtractFiles(t *testing.T) {
...
+			Cache:  s,
+		}
 
-	// setup cache
-	tempCacheDir, _ := ioutil.TempDir("", "TestDockerExtractor_SaveLocalImage-*")
-	defer func() {
-		_ = os.RemoveAll(tempCacheDir)
-	}()
+		r, err := de.SaveLocalImage(context.TODO(), "fooimage")
+		assert.NotNil(t, r, tc.name)
```
knqyf263: It's better to check the content of `r`. e.g. ioutil.ReadAll

extractor/docker/docker_test.go
```diff
@@ -187,35 +208,70 @@ func TestExtractFiles(t *testing.T) {
...
+			_ = f.Close()
+			_ = os.RemoveAll(f.Name())
+		}()
+		assert.NoError(t, err)
+
+		if tc.cacheHit {
+			_ = s.Set(kvtypes.SetItemInput{
+				BucketName: "imagebucket",
+				Key:        "fooimage",
+				Value:      []byte("foocontent"),
```
knqyf263: IMHO, we should change this value because we can't determine if the cache was hit in `SaveLocalImage`

extractor/docker/docker_test.go
```diff
@@ -13,13 +17,30 @@ import (
...
 	"github.com/stretchr/testify/assert"
 )
 
+func setupCache() (*bolt.Store, *os.File, error) {
+	f, err := ioutil.TempFile(".", "Bolt_TestStore-*")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	s, err := bolt.NewStore(bolt.Options{
```
knqyf263: In the point of unit test, we should prepare another Store like MemoryStore. 
This will be helpful. https://github.com/spf13/afero#step-2-declare-a-backend

extractor/docker/docker_test.go
```diff
@@ -338,3 +396,209 @@ func TestDockerExtractor_Extract(t *testing.T) {
...
+					}
+				}
+				assert.Equal(t, len(tc.requiredFiles), numRequiredFilesFound, tc.name)
+			}
+
+		case false:
+			assert.False(t, found, tc.name)
+			assert.Empty(t, actualCacheContents, tc.name)
+		}
+
```
knqyf263: empty line

extractor/docker/docker_test.go
```diff
@@ -338,3 +396,209 @@ func TestDockerExtractor_Extract(t *testing.T) {
...
+	}
+
```
knqyf263: ditto

extractor/docker/docker_test.go
```diff
@@ -338,3 +396,209 @@ func TestDockerExtractor_Extract(t *testing.T) {
...
+			Key:        string(inputDigest),
+			Value:      &actualCacheContents,
+		})
+
+		switch tc.cacheWB {
+		case true:
+			assert.True(t, found, tc.name)
+			assert.NoError(t, err, tc.name)
+
+			tbuf, err := extractTarContent(actualCacheContents)
```
knqyf263: If a unit test depends on the application code, it can be a meaningless test. How about preparing a zstd file by manual and compare the result with it?

extractor/docker/docker.go
```diff
@@ -196,14 +235,14 @@ func (d Extractor) Extract(ctx context.Context, imageName string, filenames []st
...
 		}(ref.Digest)
 	}
 
-	filesInLayers := make(map[string]extractor.FileMap)
-	opqInLayers := make(map[string]extractor.OPQDirs)
+	filesInLayers := map[string]extractor.FileMap{}
+	opqInLayers := map[string]extractor.OPQDirs{}
 	for i := 0; i < len(m.Manifest.Layers); i++ {
-		if err := d.extractLayerFiles(layerCh, errCh, ctx, filenames, filesInLayers, opqInLayers); err != nil {
+		if filesInLayers, opqInLayers, err = d.extractLayerFiles(layerCh, errCh, ctx, filenames); err != nil {
```
knqyf263: MUST: `extractLayerFiles` always overwrite `filesInLayers` by accident. This is a cause of a bug. `filesInLayers` must have necessary files for each layer. The current implementation has files only for one layer since it overwrites the variable.

- sha256:976a760c94fcdd7d105269ae621e8269e7bb25a58c52ae667b4029a6bc7e33cb
  - `etc/lsb-release` : content
- sha256:c58992f3c37bb64aeba18910408cda9a7a63e212fe27e95065a8d54130ca5926
  - `etc/os-release` : content

This bug affects all OSes, not just Ubuntu. It reproduces when the image has some layers. The image having only one layer doesn't overwrite the variable, so this bug doesn't reproduce. Coincidentally Ubuntu made us notice this bug 😅 We have to add this case to the unittest.

extractor/docker/docker.go
```diff
@@ -237,49 +276,170 @@ func downloadConfigFile(ctx context.Context, r *registry.Registry, image registr
...
+
+				_, err = twc.Write(fileBuf.Next(int(n)))
+				if err != nil {
+					return cacheBuf, requiredFileFound, xerrors.Errorf("%s: %s", ErrFailedCacheWrite, err)
+				}
+				break
+			}
+		}
+	}
+	if err := twc.Close(); err != nil {
```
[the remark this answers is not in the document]
simar7: Good point!

extractor/docker/docker.go
```diff
@@ -309,7 +469,7 @@ func (d Extractor) ExtractFromFile(ctx context.Context, r io.Reader, filenames [
 			break
 		}
 		if err != nil {
-			return nil, xerrors.Errorf("failed to extract the archive: %w", err)
+			return nil, xerrors.Errorf(">>> 459: failed to extract the archive: %w", err)
```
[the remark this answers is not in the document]
simar7: Just skid marks for debugging 😅 they need to be removed.

extractor/docker/docker_test.go
simar7: Yeah I see your point. Although I was being over cautious here and trying to count all files the required files to be present in the cache by un-taring the cache contents and inspecting file by file. Maybe this is bit of an overkill. We could just manually store the golden file on disk and compare the two.

extractor/docker/docker.go
simar7: Wonderful observation! Thanks. I've fixed it here: https://github.com/aquasecurity/fanal/pull/51/commits/cad6d3165f66009286fb0548498231cd4017ff94

extractor/docker/docker.go
simar7: Yeah I agree. Maybe we should add a TODO here and leave it for a future improvement as this PR is already getting big. What do you think?

extractor/docker/docker_test.go
simar7: I'll add a comment for future improvement to this.

extractor/docker/docker.go
simar7: What do you think about this change?

```diff
--git a/extractor/docker/docker.go b/extractor/docker/docker.go
index bfdd958..10d5f46 100644
--- a/extractor/docker/docker.go
+++ b/extractor/docker/docker.go
@@ -186,14 +186,14 @@ func (d Extractor) SaveLocalImage(ctx context.Context, imageName string) (io.Rea
 		found = false
 	}
 
-	var savedImage []byte
+	var savedImage bytes.Buffer
 	if err != nil || !found {
 		storedReader, err = d.saveLocalImage(ctx, imageName)
 		if err != nil {
 			return nil, xerrors.Errorf("failed to save the image: %w", err)
 		}
 
-		savedImage, err = ioutil.ReadAll(storedReader)
+		_, err := io.Copy(&savedImage, storedReader)
 		if err != nil {
 			return nil, xerrors.Errorf("failed to read saved image: %w", err)
 		}
@@ -203,7 +203,7 @@ func (d Extractor) SaveLocalImage(ctx context.Context, imageName string) (io.Rea
 			return nil, err
 		}
 
-		dst := e.EncodeAll(savedImage, nil)
+		dst := e.EncodeAll(savedImage.Bytes(), nil)
 		if err := d.Cache.BatchSet(kvtypes.BatchSetItemInput{
 			BucketName: "imagebucket",
 			Keys:       []string{imageName},
@@ -213,7 +213,7 @@ func (d Extractor) SaveLocalImage(ctx context.Context, imageName string) (io.Rea
 		}
 	}
 
-	return bytes.NewReader(savedImage), nil
+	return bytes.NewReader(savedImage.Bytes()), nil
 }
 
 func (d Extractor) saveLocalImage(ctx context.Context, imageName string) (io.ReadCloser, error) {
```

extractor/docker/docker_test.go
knqyf263: I agree.

extractor/docker/docker.go
knqyf263: I'm not sure, but it looks the same because the destination of `io.Copy` is bytes.Buffer. After all, it has all bytes in memory. To be honest, I don't have a way to sort it out now 😄

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break // end of archive
+		}
+		if err != nil {
+			return cacheBuf, false, xerrors.Errorf("%s: invalid tar: %w", ErrFailedCacheWrite, err)
+		}
+		for _, file := range requiredFilenames {
+			if file == hdr.Name {
```
knqyf263: You can use `utils.StringInSlice`.
```
if !utils.StringInSlice(hdr.Name, requiredFileNames) {
    continue
}
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+
+	tarContent, err = dec.DecodeAll(cacheContent, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return tarContent, nil
+}
+
+func getFilteredTarballBuffer(tr *tar.Reader, requiredFilenames []string) (bytes.Buffer, bool, error) {
```
knqyf263:
```suggestion
func getFilteredTarballBuffer(tr *tar.Reader, requiredFilenames []string) (bytes.Buffer, error) {
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+
+				_, err := io.Copy(twc, tr)
+				if err != nil {
+					return cacheBuf, requiredFileFound, xerrors.Errorf("%s: %s", ErrFailedCacheWrite, err)
+				}
+				break
+			}
+		}
+	}
+	return cacheBuf, requiredFileFound, nil
```
knqyf263:
```suggestion
	return cacheBuf, nil
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+		if len(filenames) > 0 {
+			if cacheBuf, storeInCache, err = getFilteredTarballBuffer(tarReader, filenames); err != nil {
+				errCh <- err
+				return
+			}
+		}
+
+		// only store in cache if a required file was found
+		// TODO: should we error out in case of a cache store failure?
+		if storeInCache {
```
knqyf263: We need to store a key with empty value into the cache even if the layer doesn't have the necessary files. If there is no cache with the layer id, `fanal` needs to fetch the layer again next time.

```suggestion
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+		// read the incoming gzip from the layer
+		gzipReader, err := gzip.NewReader(rc)
 		if err != nil {
-			log.Print(err)
+			errCh <- xerrors.Errorf("could not init gzip reader: %w", err)
+			return
+		}
+		defer gzipReader.Close()
+
+		tarReader := tar.NewReader(io.TeeReader(gzipReader, &tarContent))
```
knqyf263: `cacheBuf` should be returned as `tarContent`, isn't it? If we use `io.TeeReader`, it seems that the raw tar content will be returned.

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+
+		// only store in cache if a required file was found
+		// TODO: should we error out in case of a cache store failure?
+		if storeInCache {
+			d.storeLayerInCache(cacheBuf, dig)
 		}
 	}
-	gzipReader, err := gzip.NewReader(rc)
+
+	layerCh <- layer{ID: dig, Content: ioutil.NopCloser(bytes.NewReader(tarContent.Bytes()))}
```
knqyf263:
```suggestion
	layerCh <- layer{ID: dig, Content: ioutil.NopCloser(&tarContent)}
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+			return
+		}
+		defer gzipReader.Close()
+
+		tarReader := tar.NewReader(io.TeeReader(gzipReader, &tarContent))
+
+		var storeInCache bool
+		var cacheBuf bytes.Buffer
+		if len(filenames) > 0 {
+			if cacheBuf, storeInCache, err = getFilteredTarballBuffer(tarReader, filenames); err != nil {
```
knqyf263:
```suggestion
			if tarContent, storeInCache, err = getFilteredTarballBuffer(tarReader, filenames); err != nil {
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+				break
+			}
+		}
+	}
+	return cacheBuf, requiredFileFound, nil
+}
+
+func (d Extractor) storeLayerInCache(cacheBuf bytes.Buffer, dig digest.Digest) {
+	// compress tar to zstd before storing to cache
+	e, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(zstd.SpeedDefault))
```
knqyf263: cacheBuf is a bit large, so it may be better to use io.Copy.
https://github.com/klauspost/compress/tree/master/zstd#usage

>Smaller encodes are encouraged to use the EncodeAll function.

```suggestion
	var output bytes.Buffer
	e, err := zstd.NewWriter(output, zstd.WithEncoderLevel(zstd.SpeedDefault))
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+	if err != nil {
+		log.Println(err)
 		return
 	}
-	layerCh <- layer{ID: dig, Content: gzipReader}
+	dst := e.EncodeAll(cacheBuf.Bytes(), nil)
```
knqyf263:
```suggestion
	 _, err := io.Copy(w, &cacheBuf)
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,138 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+	if err := d.Cache.BatchSet(kvtypes.BatchSetItemInput{
+		BucketName: "layertars",
+		Keys:       []string{string(dig)},
+		Values:     dst,
```
knqyf263:
```suggestion
		Values:     output.Bytes(),
```

extractor/docker/docker_test.go
```diff
@@ -187,35 +206,79 @@ func TestExtractFiles(t *testing.T) {
...
+		}
 
-	// setup cache
-	tempCacheDir, _ := ioutil.TempDir("", "TestDockerExtractor_SaveLocalImage-*")
-	defer func() {
-		_ = os.RemoveAll(tempCacheDir)
-	}()
+		r, err := de.SaveLocalImage(context.TODO(), "fooimage")
+		actualSavedTarBytes, _ := ioutil.ReadAll(r)
+		assert.Equal(t, []byte("foo"), actualSavedTarBytes[:], tc.name)
```
knqyf263: I'm sorry my explanation may not be enough. Let me explain again, please. How do we determine if this value "foo" is from `httptest` or a cache? Even if a cache isn't hit, the value looks the same "foo". In short, how do we know this option is effective?
https://github.com/aquasecurity/fanal/pull/51/files#diff-4d409e1b4160e28284809dc87cfddec4R218

extractor/docker/docker.go
simar7: Yeah I realized that after I made this comment 😄 I was going to play with `io.Pipe` like this https://garbagecollected.org/2015/05/30/io-with-go-io-pipe/ thoughts?

extractor/docker/docker_test.go
simar7: I understand now. I have added a commit to differentiate the expectations https://github.com/aquasecurity/fanal/pull/51/commits/5f894c996176f0bfb2faa6f5be81338a7f179e10

extractor/docker/docker.go
```diff
@@ -75,13 +83,26 @@ func NewDockerExtractor(option types.DockerOption) (Extractor, error) {
...
 		Client: cli,
-		Cache:  cache.Initialize(utils.CacheDir()),
+		Cache:  kv,
 	}, nil
 }
 
+func NewDockerExtractor(option types.DockerOption) (Extractor, error) {
+	return NewDockerExtractorWithCache(option, bolt.Options{
+		RootBucketName: "fanal",
+		Path:           utils.CacheDir() + "/cache.db",
```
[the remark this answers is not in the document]
simar7: OK I'll add a TODO to revisit in a future PR.

cmd/fanal/main.go
```diff
@@ -46,10 +46,8 @@ func run() (err error) {
 	clearCache := flag.Bool("clear", false, "clear cache")
 	flag.Parse()
 
-	c := cache.Initialize(utils.CacheDir())
-
 	if *clearCache {
-		if err = c.Clear(); err != nil {
+		if err = os.RemoveAll(utils.CacheDir() + "/cache.db"); err != nil {
```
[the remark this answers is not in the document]
simar7: Hmm... that means we have to bring back the now removed ```Cache``` interface. I'm fine with bringing it back but I'd like to know if Trivy is using any of the other methods from the interface, namely: Get and Set.

extractor/docker/docker.go
simar7: Will do, thanks!

extractor/docker/docker.go
simar7: Interesting. So if a layer is stored with an empty value, fanal will retrieve an empty value when the cache is hit. Is that ok?

extractor/docker/docker.go
simar7: Yes but that's what we're after right? After passing the tar to content the Extract method will be called which expects a tar file https://github.com/aquasecurity/fanal/blob/5f894c996176f0bfb2faa6f5be81338a7f179e10/extractor/docker/docker.go#L531-L535

Maybe I misunderstood what you meant to say. Do you have a diff or a test case in mind?

extractor/docker/docker.go
knqyf263: Exactly.

knqyf263: Retrieving an empty value, fanal can know the layer is already fetched and contains no required files.

cmd/fanal/main.go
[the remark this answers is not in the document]
knqyf263: No. It's only `cache.Clear()`. If trivy can clear the cache without being aware of the path, other way may be fine as well.

extractor/docker/docker.go
knqyf263: I think storeInCache is unnecessary.
```suggestion
			if cacheBuf, err = getFilteredTarballBuffer(tarReader, filenames); err != nil {
```

extractor/docker/docker.go
```diff
@@ -257,29 +311,135 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+
+func (d Extractor) storeLayerInCache(cacheBuf bytes.Buffer, dig digest.Digest) {
+	// compress tar to zstd before storing to cache
+	var dst bytes.Buffer
+	w, _ := zstd.NewWriter(&dst)
+	_, _ = io.Copy(w, &cacheBuf)
+	_ = w.Close()
+
+	if err := d.Cache.BatchSet(kvtypes.BatchSetItemInput{
+		BucketName: "layertars",
```
knqyf263: I feel const is better.

extractor/docker/docker.go
```diff
@@ -257,29 +311,135 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+		Keys:       []string{string(dig)},
+		Values:     dst.Bytes(),
+	}); err != nil {
+		log.Printf("an error occurred while caching: %s\n", err)
```
knqyf263: nit: `log.Printf` adds `\n` automatically, so we don't have to add `\n`.

cmd/fanal/main.go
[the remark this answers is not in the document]
simar7: Thanks. I've brought it back in limited capacity: https://github.com/aquasecurity/fanal/pull/51/commits/77e25641e8faf3a6bfdacb7fb39b4a8cfda1bb6b

extractor/docker/docker.go
simar7: I see. Thanks that makes sense. I have fixed this here: https://github.com/aquasecurity/fanal/pull/51/commits/745ac0cbcb8105744323beef89cddf7d5e9449ef

extractor/docker/docker.go
```diff
@@ -257,29 +312,127 @@ func (d Extractor) extractLayerFiles(layerCh chan layer, errCh chan error, ctx c
...
+			if cacheBuf, err = getFilteredTarballBuffer(tarReader, filenames); err != nil {
+				errCh <- err
+				return
+			}
 		}
+
+		d.storeLayerInCache(cacheBuf, dig)
+	}
+
+	layerCh <- layer{ID: dig, Content: ioutil.NopCloser(&cacheBuf)}
```
knqyf263: When the layer id hits the cache, will `Content` be empty? This is because we wrote a cache content to `tarContent`. It seems to introduce a bug. I'm sorry if I misunderstand something.
https://github.com/aquasecurity/fanal/pull/51/files#diff-1ae53d6fc729587121b603c3ace40f24R330

simar7: Yes, you're correct. This indeed is a bug and was a result because of a weak test assertion: https://github.com/aquasecurity/fanal/pull/51/files/f22c0ed94fb198ad800a480303e563615458d7bc#diff-4d409e1b4160e28284809dc87cfddec4R537 I have fixed this here: https://github.com/aquasecurity/fanal/pull/51/commits/8e233b311ac7b1ced61a64346b31ba1ef952b039

extractor/docker/docker_test.go
knqyf263: I looked into the commit, but it looks the same between cached and http response data because `expectedImageData` is used here.
https://github.com/aquasecurity/fanal/commit/5f894c996176f0bfb2faa6f5be81338a7f179e10#diff-4d409e1b4160e28284809dc87cfddec4R230

In "happy path with cache hit" test, the cache data and http response are the same value, "foofromcache". We can't know if `SaveLocalImage` uses the cache or http response. 

We need to fix those values like the following

```
			case strings.Contains(httpPath, "images/get?names=fooimage"):
				_, _ = fmt.Fprint(w, "foofromdocker")
```

```
			dst := e.EncodeAll([]byte("foofromcache), nil)
```

If `SaveLocalImage` doesn't hit the cache, it returns `foofromdocker` and the test will fail because this test expect `foofromcache`. If it returns `foofromcache` as we expect, the test should succeed. Does it make sense? Feel free to ask me if you have a question.

simar7: Yeah I understand your point. I didn't realize I had factored out the expectation to be as the actual content being served 😞 I've fixed it here: https://github.com/aquasecurity/fanal/pull/51/commits/122ee971b01e351026ecacf41f70c101c970f99e

simar7:
In the interest of shipping fast and slicing thin, I'd like to merge this PR and defer improvements to future PRs.

PR #54, opened 2019-12-04
Using bufio reader for Stdin

samifruit514:
`cat out.tar | go run cmd/fanal/main.go -f -`

On my mac book, I get invalid tar header
 
otherwise the first 3 bytes are consumed by IsGzip and the file gets "corrupted"

Stdin is probably not seekable.


simar7:
Hello @samifruit514 , thanks for filling this issue. I'm trying to reproduce it locally but without luck so far.

A few follow up questions:

* Could you share the tar file here? 
* How was this tar generated? Did you by any chance use git-archive to generate this tar?

samifruit514:
Thanks for the quick feedback! Any tar won't work, as long as you send it through the stdin: you can use docker save command:
`docker save alpine | go run cmd/fanal/main.go
2019/12/04 19:55:03 failed to extract files from tar: failed to extract the archive: archive/tar: invalid tar header
exit status 1`

simar7:
Thanks @samifruit514. Sorry I didn't realize that I was on the wrong branch when I was trying to reproduce this. 

I'm able to do so now with the steps you've described. I've added some tests for the code that was modified. Once the CI run finishes successfully, I'll merge this in.

Thanks again for your contribution!

samifruit514:
thanks!

PR #66, opened 2020-01-07
integration: Add a test to use fanal as a library in Docker mode

simar7:
Addresses: #74


integration/library_test.go
```diff
@@ -0,0 +1,121 @@
...
+	_ "github.com/aquasecurity/fanal/analyzer/pkg/apk"
+	_ "github.com/aquasecurity/fanal/analyzer/pkg/dpkg"
+	_ "github.com/aquasecurity/fanal/analyzer/pkg/rpm"
+
+	"github.com/aquasecurity/fanal/cache"
+	"github.com/aquasecurity/fanal/types"
+)
+
+func TestFanal_Library_DockerMode(t *testing.T) {
+	ctx := context.Background()
```
knqyf263: First of all, we should do a table driven test because we need to test all OSes.

integration/library_test.go
```diff
@@ -0,0 +1,121 @@
...
+		Timeout:  600 * time.Second,
+		SkipPing: true,
+	}
+
+	ext, err := docker.NewDockerExtractor(opt, c)
+	require.NoError(t, err)
+	ac := analyzer.Config{Extractor: ext}
+
+	expectedFiles := []string{"etc/alpine-release", "etc/os-release", "lib/apk/db/installed", "/config"}
+	actualFiles, err := ac.Analyze(ctx, "alpine")
```
knqyf263: From where is this `alpine`? If you expect this image is in Docker Engine, we should import the image into Docker Engine before running `ac.Analyze`.

simar7: Yes, that's an oversight on my part. I'll fix it.

integration/library_test.go
simar7: Yeah I had kept this open for discussion. I wasn't sure how much detail we wanted to go into with it. Shall we add a test for each OS? Or one for each OS and the variety of options like here https://github.com/aquasecurity/trivy/blob/master/integration/standalone_test.go#L31-L343?

knqyf263: Let's add a test for each OS. We don't have many options for fanal like Trivy.

integration/library_test.go
knqyf263: As we discussed [before](https://github.com/aquasecurity/trivy/pull/335/files#diff-272fbc9a7f1f9b71135958c65d0091a6R61), we need to use a unique tag. If we don't do it, `fanal` automatically fetches an image from Docker Registry. I think this test in CI should fail, but it passed by accident. I'm not sure, but it might be because `alpine` is downloaded from Docker Hub.

simar7: Yeah, I've taken care of this now.

integration/library_test.go
```diff
@@ -0,0 +1,216 @@
...
... SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "libmount", Version: "2.30.2", Release: "2.amzn2.0.4", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "nss-pem", Version: "1.0.3", Release: "5.amzn2", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "libssh2", Version: "1.4.3", Release: "12.amzn2.2", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "rpm-libs", Version: "4.11.3", Release: "25.amzn2.0.3", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "rpm-build-libs", Version: "4.11.3", Release: "25.amzn2.0.3", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "python-pycurl", Version: "7.19.0", Release: "19.amzn2.0.2", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "yum-plugin-ovl", Version: "1.1.31", Release: "46.amzn2.0.1", Epoch: 0, Arch: "noarch", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}},
```
knqyf263: I feel it is difficult to maintain it. How about saving these results as golden files?

integration/library_test.go
```diff
@@ -0,0 +1,216 @@
...
...: "", SrcEpoch: 0}, analyzer.Package{Name: "python-dateutil", Version: "1.5", Release: "7.el7", Epoch: 0, Arch: "noarch", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "libxml2-python", Version: "2.9.1", Release: "6.el7_2.3", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "python-gobject-base", Version: "3.22.0", Release: "1.el7_4.1", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "pkgconfig", Version: "0.27.1", Release: "4.el7", Epoch: 1, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "python-ethtool", Version: "0.8", Release: "8.el7", Epoch: 0, Arch: "x86_64", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "python-decorator", Version: "3.4.0", Release: "3.el7", Epoch: 0, Arch: "noarch", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Package{Name: "python-backports-ssl_match_hostname", Version: "3.5.0.1", Release: "1.el7", Epoch: 0, Arch: "noarch", SrcName: "", SrcVersion: "", SrcRelease: "", SrcEpoch: 0}, analyzer.Packa
```
knqyf263: Is there any reason not to use `t.Run()`? It may leak `defer` unless we use `t.Run`.

knqyf263: We can just write:
```
defer os.RemoveAll(d)
```

integration/library_test.go
simar7: I'm OK with either approach. I personally think keeping them together in code is nice because they're still version controlled anyway so we'll see a change, although it's very unlikely that a change will happen. But I can turn them into golden files if you prefer.

knqyf263:
Also, I found we should test each image twice because we need to check if our cache works well.

simar7:
Hey @knqyf263 I added golden files https://github.com/aquasecurity/fanal/pull/66/commits/b42456393fc3d76e8d95d63f7e19e82e7ec21cf3, running all tests in parallel https://github.com/aquasecurity/fanal/pull/66/commits/97d468f7782d0492b50b0b34413fcf6f94d8ded2 and running each test twice https://github.com/aquasecurity/fanal/pull/66/commits/4ece53c6587a9b2434fdfcbbb93ad7946e1cdb13 support.  Have a look!

integration/library_test.go
```diff
@@ -0,0 +1,210 @@
...
+			expectedLibraries: map[analyzer.FilePath][]godeptypes.Library{},
+		},
+		{
+			name:              "happy path, opensuse leap 15.1",
+			imageName:         "opensuse/leap:latest",
+			imageFile:         "testdata/fixtures/opensuse-leap-151.tar.gz",
+			expectedFiles:     []string{"usr/lib/os-release", "usr/lib/sysimage/rpm/Packages", "/config", "etc/os-release"},
+			expectedOS:        analyzer.OS{Name: "15.1", Family: "opensuse.leap"},
+			expectedLibraries: map[analyzer.FilePath][]godeptypes.Library{},
+		},
```
knqyf263: Would you add the following image? The image has libraries. We should test a case when `expectedLibraries` is not empty.
https://github.com/aquasecurity/trivy-test-images/blob/master/vulnimage.tar.gz

simar7:
hey @knqyf263 I fixed most of the stuff but I'm unable to find a docker engine version that is compatible with the client we're using. Would you mind looking into it? We're using 1.39 and the docker-engine on both xenial and bionic is only compatible up to 1.38.

knqyf263:
@simar7 Really? It's weird. It means Trivy doesn't work on Ubuntu in Docker mode. We have to investigate it more.

simar7:
OK @knqyf263 I took a look at it, seems that the version of docker that comes with the default travis image is fairly old. Once I upgraded to the latest stable version of docker that ships with Ubuntu bionic, everything was fine. Here's the change https://github.com/aquasecurity/fanal/pull/66/commits/035c2a99b5d522d5ee51ec5597d39ec432d952fd

simar7:
I think we should merge this now, if you're happy with the changes.

knqyf263:
@simar7 I see. Thank you for your investigation!

Issue #58, opened 2019-12-16
Improve the scan speed in Docker mode


knqyf263:
https://github.com/aquasecurity/fanal/pull/51 made Docker mode slower.

Before
```
$ /usr/bin/time ./fanal centos:7
{Name:7.6.1810 Family:centos}
via image Packages: 146
via file Packages: 0
        0.22 real         0.17 user         0.13 sys
```

After
```
$ /usr/bin/time ./fanal centos:7
{Name:7.6.1810 Family:centos}
via image Packages: 146
via file Packages: 0
        1.34 real         1.24 user         0.13 sys
```

PR #67, opened 2020-01-10
Revert PR #51 entirely

simar7:
Currently caching introduces extra memory pressure which is causing issues like https://github.com/aquasecurity/trivy/issues/362 

This PR reverts it.


extractor/docker/docker.go
```diff
@@ -73,7 +63,7 @@ type Extractor struct {
 	cache  cache.Cache
 }
 
-func NewDockerExtractor(option types.DockerOption, cache cache.Cache) (Extractor, error) {
+func NewDockerExtractor(option types.DockerOption) (Extractor, error) {
```
knqyf263: You should keep `cache.Cache` the argument. If we remove it, `cache.Initialize` always take `utils.CacheDir`. 
https://github.com/aquasecurity/fanal/pull/67/files#diff-1ae53d6fc729587121b603c3ace40f24R78

But Trivy needs to specify the cache directory from a CLI option. Also, I want to avoid changing the interface.

extractor/docker/docker.go
```diff
@@ -203,7 +168,7 @@ func (d Extractor) saveLocalImage(ctx context.Context, imageName string) (io.Rea
 }
 
 func (d Extractor) Extract(ctx context.Context, imageName string, filenames []string) (extractor.FileMap, error) {
-	ctx, cancel := context.WithTimeout(ctx, d.Option.Timeout)
+	ctx, cancel := context.WithTimeout(context.Background(), d.Option.Timeout)
```
knqyf263: You don't have to revert it as well. This is like a bug. We should use `ctx`.

extractor/docker/docker.go
```diff
@@ -232,10 +197,10 @@ func (d Extractor) Extract(ctx context.Context, imageName string, filenames []st
 		}(ref.Digest)
 	}
 
-	filesInLayers := map[string]extractor.FileMap{}
-	opqInLayers := map[string]extractor.OPQDirs{}
+	filesInLayers := make(map[string]extractor.FileMap)
+	opqInLayers := make(map[string]extractor.OPQDirs)
 	for i := 0; i < len(m.Manifest.Layers); i++ {
-		if err := d.extractLayerFiles(ctx, layerCh, errCh, filesInLayers, opqInLayers, filenames); err != nil {
+		if err := d.extractLayerFiles(layerCh, errCh, ctx, filenames, filesInLayers, opqInLayers); err != nil {
```
knqyf263: nit: I prefer `ctx` is the first argument.

cache/cache.go
```diff
@@ -1,65 +1,58 @@
...
-		return nil, xerrors.Errorf("unable to create cache dir: %w", err)
-	}
-
-	cacheOptions := bolt.Options{
-		RootBucketName: "fanal",
-		Path:           filepath.Join(dir, "cache.db"),
-		Codec:          encoding.Raw,
-	}
+func Initialize(cacheDir string) Cache {
+	return &FSCache{Directory: cacheDir}
```
knqyf263:
```suggestion
	return &FSCache{Directory: filepath.Join(cacheDir, "fanal")}
```

`cacheDir` is usually a system cache directory itself. So, in my environment, it is `/Users/teppei/Library/Caches/`. If we scan `alpine:3.10`, it will be stored in `/Users/teppei/Library/Caches/alpine:3.10`. We should add one more directory to store image caches. The system cache directory will be deleted with `--clear` option if we don't add one more directory.

Besides, we should define it in const like the below.

```
const(
    cacheDirName = "fanal"
)
```

cache/cache.go
```diff
@@ -1,65 +1,62 @@
...
+	Get(key string) io.Reader
+	Set(key string, file io.Reader) (io.Reader, error)
 	Clear() error
 }
 
-type RealCache struct {
-	directory string
-	cache     *bolt.Store
+type FSCache struct {
+	Directory string
```
knqyf263: Do we need to export this value? If we always pass this value via `Initialize`, we should unexport it.

```suggestion
	directory string
```

knqyf263:
Fix https://github.com/aquasecurity/fanal/issues/58 once by this revert

PR #70, opened 2020-01-15
refactor: replace genuinetools/reg with containers/image

knqyf263:
## Overview
replace [genuinetools/reg](https://github.com/genuinetools/reg) with [containers/image](https://github.com/containers/image)

## Note
To build `fanal`, `containers_image_storage_stub` tag is necessary since `containers/storage` which `containers/image` depends on uses cgo. We don't need the features that `containers/storage` provides, so we can avoid depending C libraries by adding `containers_image_storage_stub` tag.
https://github.com/containers/image/blob/master/storage/storage_image.go#L1


simar7:
Just putting it out here, I think we should merge this PR in https://github.com/aquasecurity/fanal/pull/66 before we merge this one. It will help us with the fact if making the switch breaks anything or not.

knqyf263:
@simar7 I agree 👍

analyzer/analyzer.go
```diff
@@ -118,37 +112,19 @@ func RequiredFilenames() []string {
 
 // TODO: Remove opts as they're no longer needed
 func (ac Config) Analyze(ctx context.Context, imageName string, opts ...types.DockerOption) (fileMap extractor.FileMap, err error) {
-	r, err := ac.Extractor.SaveLocalImage(ctx, imageName)
+	transports := []string{"docker-daemon:", "docker://"}
```
mrueg: This might be outside of the scope of this PR, it would be great to allow non-docker stored images here as well, so trivy could operate on images built with buildah without doing a docker conversion.
https://github.com/containers/image/blob/master/docs/containers-transports.5.md#containers-storagestorage-specifierimage-iddocker-referenceimage-id

knqyf263: Thank you for the idea! Actually, one of the reasons for using containers/image is exactly that. I'd like to support OCI and `containers-storage` in the future.

simar7:
I tried to review this PR but unfortunately it's a bit too big for me to wrap my head around so I'm still looking into it. 

@knqyf263 would you mind rebasing/cherry-picking this PR with https://github.com/aquasecurity/fanal/pull/76 and https://github.com/aquasecurity/fanal/pull/72/ to verify its functional aspects?

knqyf263:
@simar7 Sure. Most part of the added lines are from image_test.go. The other part is just replacing the existing code. You can build and run it for testing.

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+	IsFile bool
+}
+
+type Image struct {
+	name       string   // e.g. alpine:3.10
+	isFile     bool     // from a tar file
+	transports []string // e.g. "docker://"
+
+	systemContext *imageTypes.SystemContext
+	blobInfoCache imageTypes.BlobInfoCache
```
simar7: Question: What are they used for? It's not clear to me reading it right away.

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+	// ignore the cache error
+	layerInfos, _ := img.getLayerInfosInCache(img.name, img.isFile)
+	if layerInfos != nil {
+		return layerInfos, nil
+	}
+
+	// When it doesn't hit a cache, it fetches the image from Docker Engine or Docker Registry
+	if err := img.populateSource(); err != nil {
+		return nil, xerrors.Errorf("failed population: %w", err)
+	}
```
simar7: What happens if this returns an error? I think this is untested right now.

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+
+	layers := img.src.LayerInfos()
+
+	// ignore the cache error
+	_ = img.storeLayerInfosInCache(img.name, img.isFile, layers)
```
simar7: Looks like this is the only place where `img.storeLayerInfosInCache` is called. If the error is not important to handle we should remove it from the signature of the method as it is unexported anyways.

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := img.populateSource(); err != nil {
+		return nil, xerrors.Errorf("failed population: %w", err)
+	}
+
+	rc, _, err := img.rawSource.GetBlob(ctx, imageTypes.BlobInfo{Digest: dig, Size: -1}, img.blobInfoCache)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to download the layer(%s): %w", dig, err)
+	}
+
+	stream, _, err := compression.AutoDecompress(rc)
```
simar7: Do we call `Close()` on `stream`? As per the documentation we need too it seems https://github.com/containers/image/blob/master/pkg/compression/compression.go#L130-L134

We should also add tests for this. Currently there aren't any. One of the things to assert in the test would be to ensure `Close()` is eventually called as it seems to be a requirement by the dependency.

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	r, _ := img.cache.Set(dig.String(), stream)
+
+	return ioutil.NopCloser(r), nil
+}
+
+func (img Image) RecordDigestUncompressedPair(dig digest.Digest, uncompressed digest.Digest) {
+	img.blobInfoCache.RecordDigestUncompressedPair(dig, uncompressed)
+}
```
simar7: This doesn't seem to be used anywhere. Remove?

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+	return b, nil
+}
+
+func (img *Image) GetBlob(ctx context.Context, dig digest.Digest) (io.ReadCloser, error) {
+	rc := img.cache.Get(dig.String())
+	if rc != nil {
+		return rc, nil
+	}
+
+	if err := img.populateSource(); err != nil {
```
simar7: `populateSource()` doesn't seem to be unit tested on it's own nor does it seem to have a mock in a situation like this call site. We should add both as it seems to be an important function that's commonly used by various other functions.

extractor/image/image.go
```diff
@@ -0,0 +1,240 @@
...
+func (img Image) getConfigBlobInCache(imageName string, isFile bool) ([]byte, error) {
+	// tar file doesn't have info in cache
+	if isFile {
+		return nil, nil
+	}
+	rc := img.cache.Get("configblob::" + imageName)
+	if rc == nil {
+		return nil, nil
+	}
+	b, err := ioutil.ReadAll(rc)
```
simar7: How big these blobs can get? We're doing `ioutil.ReadAll()` here so should be mindful.

extractor/image/image.go
knqyf263: `imageTypes.BlobInfoCache` can be used to store a relationship between compressed and decompressed layer IDs. We sometimes need to know a decompressed layer ID for cache without decompressing a layer actually. I introduced it once, but I removed the implementation to keep this PR small. I will add it again in the next PR. If this bothers you, I can remove it for now.

extractor/image/image.go
knqyf263: OK. I will add it in the next PR.

extractor/image/image.go
knqyf263: It's around 1~2 KB.
```
$ ls -alh fanal/configblob::*
-rw-r--r-- 1 teppei 1.5K  1 26 14:42 fanal/configblob::docker.io_library_alpine:3.7
-rw-r--r-- 1 teppei 2.2K  1 26 14:50 fanal/configblob::docker.io_library_centos:7
-rw-r--r-- 1 teppei  591  1 26 14:50 fanal/configblob::gcr.io_distroless_base-debian10:latest
```

extractor/image/image.go
knqyf263: I fixed each test not to skip `populateSource`. I feel we can test this function enough via other unit tests.

extractor/image/image.go
knqyf263: I added a defer function to clean up io.ReadCloser.
https://github.com/aquasecurity/fanal/pull/70/commits/8672e91ed5f4031ac56bd7fb9fce863fc6804fd7

extractor/image/image.go
knqyf263: I think it's up to the caller to ignores the error. `storeLayerInfosInCache` takes only responsibility to store cache and returns an error. But I'm not particular about this. If you prefer not to return an error, I will fix it.

extractor/image/image.go
knqyf263: I added the test case.

extractor/image/image.go
simar7: Let's add this as part of a separate PR. Would be easier to keep track that way.

knqyf263: Sorry, we need it here.
https://github.com/aquasecurity/fanal/pull/70/files#diff-525dd35201b72b43ad63b85dd2720ba6R224

extractor/image/image.go
simar7: Looks much better now – thanks!

extractor/image/image.go
simar7: Yeah, I'm OK with it. Just thought we could remove some dead code if it will never see the light of the day. Let's keep it for now.

simar7:
lgtm!

PR #71, opened 2020-01-15
Benchmark tests with cob

simar7:
This PR adds performance tests for fanal with cob.


simar7:
Closing in interest of https://github.com/aquasecurity/fanal/compare/perf-testing-with-cob?expand=1

Issue #73, opened 2020-01-17
Integration tests in tar scan mode


knqyf263:
Add integration tests to scan a tar file

PR #76, opened 2020-01-18
integration: Fanal as a library for tar mode

simar7:
This PR adds integration tests for Fanal being used as a library in Tar mode.

Also refactors code.

Addresses: https://github.com/aquasecurity/fanal/issues/73 


integration/library_dockermode_test.go
```diff
@@ -0,0 +1,186 @@
...
+				Force:         true,
+				PruneChildren: true,
+			})
+			assert.NoError(t, err, tc.name)
+		})
+	}
+}
+
+func runChecksDockerMode(t *testing.T, ac analyzer.Config, ctx context.Context, tc testCase, d string, c cache.Cache) {
+	actualFiles, err := ac.Analyze(ctx, tc.imageFile)
```
knqyf263: discussion: The difference between tar mode and docker mode is almost only this line. What do you think about aggregating those tests?


```
for _, fromFile := []bool{false, true} {
    var actualFiles extractor.FileMap
    var err error

    if fromFile {
        testfile, err := os.Open(tc.imageFile)
        require.NoError(t, err)
        actualFiles, err = ac.AnalyzeFile(ctx, testfile)
    } else {
        actualFiles, err = ac.Analyze(ctx, tc.imageFile)
    }
}
```

I know this is not a good example because we have dockerless mode as well, but anyway I think we can remove duplicate code, especially testcases. I'm not sure if the aggregation is a better way, but I'm afraid to forget to add a new test case in either mode.

simar7: Yeah I agree. I have addressed this concern in the next commits. I think I'm happy with having a single file and a common set of use cases as seen here: https://github.com/aquasecurity/fanal/pull/76/commits/41881b073d3f0bb9ba2e57069a7e3f7fbaebcf0f

simar7: Also this way we can run docker mode and tar mode in parallel.

integration/library_test.go
```diff
@@ -211,14 +257,33 @@ func runChecks(t *testing.T, ac analyzer.Config, ctx context.Context, tc testCas
 	} else {
 		assert.Equal(t, map[analyzer.FilePath][]godeptypes.Library{}, actualLibs, tc.name)
 	}
+}
 
-	// check Cache
-	actualCachedFiles, _ := ioutil.ReadDir(d + "/fanal/")
-	require.Equal(t, 1, len(actualCachedFiles), tc.name)
+func checkPackageFromCommands(osFound analyzer.OS, actualFiles extractor.FileMap, t *testing.T, tc testCase) {
```
knqyf263: ditto
```suggestion
func checkPackageFromCommands(t *testing.T, osFound analyzer.OS, actualFiles extractor.FileMap, tc testCase) {
```

integration/library_test.go
```diff
@@ -163,42 +163,88 @@ func TestFanal_Library_DockerMode(t *testing.T) {
...
+			opt := types.DockerOption{
+				Timeout:  600 * time.Second,
+				SkipPing: true,
+			}
+
+			cli, err := client.NewClientWithOpts(client.FromEnv)
+			require.NoError(t, err, tc.name)
+
+			// ensure image doesnt already exists
+			_, _ = cli.ImageRemove(ctx, tc.imageFile, dtypes.ImageRemoveOptions{
```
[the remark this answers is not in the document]
simar7: Yes this makes the test idempotent.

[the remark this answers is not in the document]
knqyf263: This test doesn't use Docker Daemon because this is tests for tar mode. Why is this related to idempotent?

PR #79, opened 2020-01-24
integration: Fix filenames to not include the `:` char

simar7:
Only valid chars allowed: https://github.com/golang/go/blob/1fb7d5472e8f46faaa034fe6e16ca66a1e7c766f/src/cmd/go/internal/get/path.go#L151


integration/library_test.go
```diff
@@ -270,7 +270,9 @@ func checkPackageFromCommands(t *testing.T, actualFiles extractor.FileMap, osFou
 func checkPackages(actualFiles extractor.FileMap, t *testing.T, tc testCase) {
 	actualPkgs, err := analyzer.GetPackages(actualFiles)
 	require.NoError(t, err)
-	data, _ := ioutil.ReadFile(fmt.Sprintf("testdata/goldens/%s.expectedpackages.golden", strings.ReplaceAll(tc.imageName, "/", "")))
+	goldenFile := fmt.Sprintf("testdata/goldens/%s.expectedpackages.golden", strings.ReplaceAll(strings.ReplaceAll(tc.imageName, "/", ""), ":", ""))
```
knqyf263: nit: I prefer strings.Replacer https://golang.org/pkg/strings/#Replacer.Replace

simar7: thanks! I actually wanted to do something like that, didn't know about it. I will use it.

PR #84, opened 2020-02-01
feat(cache): based on JSON

knqyf263:
https://github.com/aquasecurity/trivy/issues/395


cache/cache.go
```diff
@@ -1,77 +1,167 @@
...
+			if l.SchemaVersion != types.LayerJSONSchemaVersion {
+				missingLayerIDs = append(missingLayerIDs, layerID)
+			}
+
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
```
simar7: Since we never return anything but nil, this can be removed.

integration/library_test.go
```diff
@@ -174,96 +194,77 @@ func TestFanal_Library_TarMode(t *testing.T) {
...
-func commonChecks(t *testing.T, actualFiles extractor.FileMap, tc testCase) {
-	checkFiles(t, actualFiles, tc)
-	osFound := checkOS(actualFiles, t, tc)
-	checkPackages(actualFiles, t, tc)
-	checkPackageFromCommands(t, actualFiles, osFound, tc)
-	checkLibraries(actualFiles, t, tc)
+func commonChecks(t *testing.T, detail types.ImageDetail, tc testCase) {
+	assert.Equal(t, tc.expectedOS, *detail.OS, tc.name)
+	checkPackages(detail, t, tc)
+	//checkPackageFromCommands(t, actualFiles, osFound, tc)
```
simar7: We should bring this back but before we can, we need to figure out how we can get access to `actualFiles` again. `ImageDetail` no longer has this information.

cache/cache.go
knqyf263: I thought the same thing before, but I found db.View itself sometimes returned an error. For example, it is when db is already closed.

integration/library_test.go
knqyf263: You're right. `PackageFromCommands` gets packages like the following.

```
RUN apk add sqlite-libs \
         && ./configure --with-sqlite \
         && make && make install \
         && apk del sqlite-libs
```

`sqlite-libs` is installed/removed in one layer, so we can't find it in a merged layer. But it is linked statically. Now, trivy detects vulnerabilities of `sqlite-libs` in this case.
This PR disables this feature. As you said, we have to figure out how we support this one. It makes this PR bigger. I'm supposed to implement it in a dedicated PR.

cache/cache.go
```diff
@@ -1,77 +1,221 @@
...
+	// decompressedDigestBucket stores a mapping from any digest to an uncompressed digest.
+	decompressedDigestBucket = "decompressed"
 )
 
 type Cache interface {
-	Get(key string) (reader io.ReadCloser)
-	Set(key string, file io.Reader) (reader io.Reader, err error)
-	SetBytes(key string, value []byte) (err error)
+	ImageCache
+	LocalImageCache
```
simar7: For my understanding: standalone and server can use `FsCache` which implements the full `Cache` interface and client uses `RemoteCache` which implements the `ImageCache` interface.

What's the purpose of `PutImage` and `GetImage`? 
- Another way to ask that question would be that can we re-use `PutLayer` and `GetLayer` to store the information in ImageInfo which is used in these methods. There's an overlap of some items like `SchemaVersion` and `OS` information.

knqyf263: >For my understanding: standalone and server can use FsCache which implements the full Cache interface and client uses RemoteCache which implements the ImageCache interface.

Exactly. `RemoteCache` doesn't need `GetLayer`, `GetImage` and `Clear`, so I prepared the another interface.

knqyf263: >Another way to ask that question would be that can we re-use PutLayer and GetLayer to store the information in ImageInfo which is used in these methods. There's an overlap of some items like SchemaVersion and OS information.

`PutImage` and `GetImage` is used to store image metadata in the cache. They are completely different from layer information. But I considered the same thing that we could replace `PutLayer` with `PutBlob` and reuse it for `ImageInfo` as well. But the `ImageInfo` has the different structure from `LayerInfo`. `OS` of `LayerInfo` is a result of analysis for each layer (e.g. `alpine` or `redhat`) and means OS of the image itself, but `OS` of `ImageInfo` is from Docker metadata, which indicates on which OS the image runs like `linux`, `macOS` or `windows. Actually, this is unnecessary for now. I store these metadata from Docker for the future.

If we reuse `PutBlob`, it is difficult to restrict a type. I mean the definition is like `PutBlob([]byte) or PutBlob(interface{})`. If we separate them, we can define `PutLayer(LayerInfo)` and `PutImage(ImageInfo)`. I'm not sure which is better.

simar7:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #86, opened 2020-02-14
test(integration/bench): wait for an image load and remove images after tests

knqyf263:
## Problem 1
After `cli.ImageLoad`, it takes time for the image to become available. If you access the docker image just after `cli.ImageLoad`, it sometimes returns `no such image`. We need to wait for finishing an image load.

## Problem 2
GitHub Actions has 14 GB disk space, but it sometimes exceeds it. This PR removes images/cache every test.

```
Each virtual machine has the same hardware resources available.
- 2-core CPU
- 7 GB of RAM memory
- 14 GB of SSD disk space
https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
```


simar7:
Nice find! lgtm!

PR #87, opened 2020-02-26
Add layer id info

simar7:
This PR adds LayerID to Package struct.


analyzer/analyzer.go
```diff
@@ -176,6 +176,7 @@ func (ac Config) analyzeLayer(ctx context.Context, dig digest.Digest) (digest.Di
 	}
 
 	layerInfo := types.LayerInfo{
+		ID:            dig,
```
knqyf263: I think we don't need to store `ID` here because a cache key already has that information. We can reduce the cache size.

analyzer/analyzer.go
```diff
@@ -234,6 +235,10 @@ func (a Applier) ApplyLayers(imageID digest.Digest, layerIDs []string) (types.Im
 		if layer.SchemaVersion == 0 {
 			return types.ImageDetail{}, xerrors.Errorf("layer cache missing: %s", layerID)
 		}
+		// TODO: Should we handle this?
+		//if layer.ID == "" {
+		//	layer.ID = imageID
+		//}
```
knqyf263: I think we should insert `ID` here.
```suggestion
		layer.ID = layerID
		
```

extractor/docker/docker.go
```diff
@@ -98,7 +98,15 @@ func ApplyLayers(layers []types.LayerInfo) types.ImageDetail {
...
-			nestedMap.SetByString(pkgInfo.FilePath, sep, pkgInfo)
+			var pkgs []types.Package
+			for _, pkg := range pkgInfo.Packages {
+				pkg.LayerID = layer.ID
+				pkgs = append(pkgs, pkg)
+			}
+			nestedMap.SetByString(pkgInfo.FilePath, sep, types.PackageInfo{
+				FilePath: pkgInfo.FilePath,
+				Packages: pkgs,
+			})
```
knqyf263: nit: you can also write like the following to save memory and lines

```suggestion
			for i, pkg := range pkgInfo.Packages {
				pkgInfo.Packages[i].LayerID = layer.ID
			}
			nestedMap.SetByString(pkgInfo.FilePath, sep, pkgInfo)
```
But I like your way from a readability perspective.

analyzer/analyzer.go
simar7: Good point. I will remove this.

extractor/docker/docker.go
simar7: I actually like your proposed way better. Much cleaner. I changed to that.

analyzer/analyzer_test.go
```diff
@@ -255,6 +255,7 @@ func TestApplier_ApplyLayers(t *testing.T) {
 					},
 					Returns: cache.LocalImageCacheGetLayerReturns{
 						LayerInfo: types.LayerInfo{
+							ID:            "sha256:932da51564135c98a49a34a193d6cd363d8fa4184d957fde16c9d8527b3f3b02",
```
knqyf263: `GetLayer` doesn't return `ID`, right? We removed insertion of `ID` to the cache.
https://github.com/aquasecurity/fanal/pull/87/files/3091476f8b8234e52a44ad799d69209305f2d6ea#r384901905

analyzer/analyzer_test.go
```diff
@@ -347,13 +350,20 @@ func TestApplier_ApplyLayers(t *testing.T) {
...
+						Name: "tzdata", Version: "2019a-0+deb9u1", SrcName: "tzdata", SrcVersion: "2019a-0+deb9u1",
+						LayerID: "sha256:932da51564135c98a49a34a193d6cd363d8fa4184d957fde16c9d8527b3f3b02",
+					},
 				},
 				Applications: []types.Application{
 					{
 						Type: "composer", FilePath: "php-app/composer.lock",
-						Libraries: []depTypes.Library{{Name: "guzzlehttp/guzzle", Version: "6.2.0"},
+						Libraries: []depTypes.Library{
+							{Name: "guzzlehttp/guzzle", Version: "6.2.0"},
```
knqyf263: I'm sorry not to tell you. We also need to add `LayerID` to `Library`. A vulnerability is detected from `LIbrary` as well as `Package`. To show which layer this vulnerability comes from, we require `LayerID` in `Library`.

extractor/docker/docker.go
```diff
@@ -98,6 +98,9 @@ func ApplyLayers(layers []types.LayerInfo) types.ImageDetail {
 		}
 
 		for _, pkgInfo := range layer.PackageInfos {
+			for i := range pkgInfo.Packages {
+				pkgInfo.Packages[i].LayerID = layer.ID
+			}
 			nestedMap.SetByString(pkgInfo.FilePath, sep, pkgInfo)
 		}
 		for _, app := range layer.Applications {
```
knqyf263: Would you do the same thing for `Applications[].Libraries` as `PackagesInfos[].Packages` here?

```
			for i, libs := range app.Libraries {
				app[i].LayerID = layer.ID
			}
```

analyzer/analyzer.go
```diff
@@ -234,6 +234,9 @@ func (a Applier) ApplyLayers(imageID digest.Digest, layerIDs []string) (types.Im
 		if layer.SchemaVersion == 0 {
 			return types.ImageDetail{}, xerrors.Errorf("layer cache missing: %s", layerID)
 		}
+		if layer.ID == "" {
+			layer.ID = digest.Digest(layerID)
+		}
```
knqyf263:
```suggestion
		layer.ID = digest.Digest(layerID)
```
`layer.ID` should be always empty. We don't need this `if` condition. Even if it is not empty (it is not supposed to happen), we should overwrite `layer.ID` with `layerID`.

knqyf263:
@simar7 Also, you probably have to fix integration tests as well.

analyzer/analyzer_test.go
simar7: Yes sorry – cruft.

analyzer/analyzer_test.go
simar7: Just to confirm the Library struct comes from this package? https://github.com/aquasecurity/go-dep-parser/blob/master/pkg/types/types.go 

If so we'll have to add it there. Sidenote: turns out that package as lots of un-named struct initializers so adding a new field will require fixing all of them 😞

knqyf263: Ah, I forgot `Package` came from `go-dep-parser`. Hmm... `go-dep-parser` is a library not related to a container. So I'd like to avoid adding `LayerID` to `Package`. Do you have any idea? Our goal is to display a layer ID with a vulnerability.

simar7: I went with creating a new struct and encapsulating the Library struct inside of it.

simar7:
Bizarrely enough the unknown manifest problem seems to have returned @knqyf263 what did you do last time to fix it? Was it ensuring images were being loaded?

knqyf263:
@simar7 Yes, it sometimes hasn't finished loading the image, so I waited for it.

simar7:
I have managed to figure out what's causing this issue. It looks like we're running out of disk space inside the CI worker. 
image

This issue also does highlight another fact that if a  transport fails, fanal will silently try to use the next transport https://github.com/aquasecurity/fanal/blob/master/extractor/image/image.go#L109 . In this case the `docker:daemon:` transport fails so it tries `docker://` but this isn't obvious as we only report the last error: https://github.com/aquasecurity/fanal/blob/master/extractor/image/image.go#L123 We should probably improve this to save us time in the future for debugging.


Another issue I also found that Github Actions workers by default ship with around `13` images. I'm not sure what these images are for but they do take up space away from our use. 
image 
This is why loading a small image like `alpine:3.10` works fine but `amazonlinux:2` breaks. Here's the code that does the `io.Copy` to a tempdir  https://github.com/containers/image/blob/27e4b201067a70dba8eb58f903f52530811ca0ef/docker/tarfile/src.go#L116-L122

knqyf263:
@simar7 Yes, I'd like to improve the problem, so I thought of adding a debug message. But `fanal` is a library, so if it displays log messages, it will be mixed with log messages from an application importing `fanal`. The best option might be to pass the logger to `fanal` from outside. Do you have any idea?

>Another issue I also found that Github Actions workers by default ship with around 13 images.
What kind of images are they? If it is not necessary for us, we should remove them before running the benchmarks.

simar7:
I'm going to remove other OSes for the timebeing as fixing performance tests are not related to this story. We'll come back to them later.

analyzer/library/bundler/bundler.go
```diff
@@ -37,7 +36,11 @@ func (a bundlerLibraryAnalyzer) Analyze(fileMap extractor.FileMap) (map[types.Fi
 		if err != nil {
 			return nil, xerrors.Errorf("error with %s: %w", filename, err)
 		}
-		libMap[types.FilePath(filename)] = libs
+		for _, lib := range libs {
+			libMap[types.FilePath(filename)] = append(libMap[types.FilePath(filename)], types.LibraryInfo{
+				Library: lib,
+			})
+		}
```
knqyf263: I feel we should fix it here.
https://github.com/aquasecurity/fanal/blob/e6c70ea926f2d03f8f40b8d85023eef6dd1ef2c8/analyzer/analyzer.go#L319

analyzer/library/cargo/cargo.go
```diff
@@ -35,7 +34,11 @@ func (a cargoLibraryAnalyzer) Analyze(fileMap extractor.FileMap) (map[types.File
 		if err != nil {
 			return nil, xerrors.Errorf("error with %s: %w", filename, err)
 		}
-		libMap[types.FilePath(filename)] = libs
+		for _, lib := range libs {
+			libMap[types.FilePath(filename)] = append(libMap[types.FilePath(filename)], types.LibraryInfo{
+				Library: lib,
+			})
+		}
```
knqyf263: ditto. Other analyzers as well.

analyzer/library/bundler/bundler.go
simar7: Good point.

knqyf263:
I've fixed a problem regarding the out of disk space. containers/image stores an image cache in `/var/tmp` until we call `Close()`. In some cases, we couldn't call `Close()` well. As a result, it used a lot of disk space. I made sure calling  `Close()` every benchmark.

PR #91, opened 2020-03-10
WIP: Show Layer ID of the origin layer

simar7:
Currently we show LayerID of the layer that has all the accumulated packages, this PR will change the behaviour to show the LayerID of the layer where the vulnerable package was first found.

### DIFF
--- a/extractor/docker/docker.go
+++ b/extractor/docker/docker.go
@@ -84,17 +84,58 @@ func newDockerExtractor(ctx context.Context, imgRef image.Reference, transports 
 	}, cleanup, nil
 }
 
+func containsPackage(s []types.Package, e types.Package) bool {
+	for _, a := range s {
+		if a.Name == e.Name && a.Version == e.Version {
+			return true
+		}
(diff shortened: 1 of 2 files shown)


knqyf263:
I pushed some tests which should be passed, but I think there are more failing cases to be supposed to pass. I feel your way should be difficult. I think we should merge all layers once, then look for the layer where the package is found. Also, you can play with Dockerfile.

For example:
```
FROM python:3.8-buster
RUN apt-get install zstd
RUN apt-get remove zstd
RUN pipenv install django==3.0
RUN rm Pipfile.lock
RUN pipenv install django==3.0.1
```
or something like that. Then, you can export them `docker save -o test.tar your_image:latest`. You will find edge cases during playing with some images. In addition, I recommend reading the specification.
https://github.com/moby/moby/blob/master/image/spec/v1.2.md

simar7:
Deprecated in favour of: https://github.com/aquasecurity/fanal/pull/93

PR #95, opened 2020-03-16
feat: support stdin

knqyf263:
fanal supported `-` before, but the feature was dropped when moving to `containers/image`. But some users want this feature, so this PR makes it possible to do it again.
Related to https://github.com/aquasecurity/trivy/pull/418

### DIFF
--- a/extractor/image/image.go
+++ b/extractor/image/image.go
@@ -3,6 +3,8 @@ package image
 import (
 	"context"
 	"io"
+	"io/ioutil"
+	"os"
 
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/pkg/blobinfocache"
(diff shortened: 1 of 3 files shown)


extractor/image/image.go
```diff
@@ -69,6 +73,16 @@ func NewImage(ctx context.Context, image Reference, transports []string, option
 		auth = GetToken(ctx, domain, option)
 	}
 
+	// Receive an image content from stdin
+	var tmpImageFileName string
+	if image.IsFile && image.Name == "-" && !terminal.IsTerminal(0) {
```
simar7: Is there a way to set something up so that `!terminal.IsTerminal(0)` can evaluate to `true` in a test? If so we should add a test for it.

PR #96, opened 2020-03-23
fix: replace containers/image with google/go-containerregistry

knqyf263:
## TODO
- [x] replace containers/image with google/go-containerregistry
- [x] use diff_id as a key of cache instead of digest


cache/cache.go
```diff
@@ -95,14 +91,8 @@ func (fs FSCache) GetLayer(layerID string) (types.LayerInfo, error) {
...
-func (fs FSCache) getLayer(decompressedBucket, layerBucket *bolt.Bucket, layerID string) (types.LayerInfo, error) {
-	// get a decompressed layer ID
-	d := decompressedBucket.Get([]byte(layerID))
-	if d != nil {
-		layerID = string(d)
-	}
-
-	b := layerBucket.Get([]byte(layerID))
+func (fs FSCache) getLayer(layerBucket *bolt.Bucket, diffID string) (types.LayerInfo, error) {
+	b := layerBucket.Get([]byte(diffID))
```
simar7, line 95: Can `b` be `nil`?

cmd/fanal/main.go
```diff
@@ -67,15 +68,15 @@ func run() (err error) {
 		SkipPing: true,
 	}
 
+	cleanup := func() {}
```
simar7, line 71: What's this for?

extractor/docker/docker.go
```diff
@@ -186,41 +180,72 @@ func ApplyLayers(layers []types.LayerInfo) types.ImageDetail {
...
 		return "", nil, nil, nil, xerrors.Errorf("failed to extract files: %w", err)
 	}
 
-	decompressedLayerID := digest.NewDigestFromBytes(digest.SHA256, sha256hash.Sum(nil))
+	return digest, files, opqDirs, whFiles, nil
+}
 
-	return decompressedLayerID, files, opqDirs, whFiles, nil
+func (d Extractor) isCompressed(l v1.Layer) bool {
+	_, uncompressed := reflect.TypeOf(l).Elem().FieldByName("UncompressedLayer")
```
simar7, line 247: What if `l` is `nil`?

extractor/image/daemon/image_test.go
```diff
@@ -0,0 +1,251 @@
...
+		APIVersion: "1.38",
+		ImagePaths: imagePaths,
+	}
+	te := engine.NewDockerEngine(opt)
+	defer te.Close()
+
+	os.Setenv("DOCKER_HOST", fmt.Sprintf("tcp://%s", te.Listener.Addr().String()))
+
+	code := m.Run()
+	os.Exit(code)
```
simar7, line 35:
```suggestion
	os.Exit(m.Run())
```

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		defer rc.Close()
+
+		if _, err = io.Copy(f, rc); err != nil {
+			return nil, xerrors.Errorf("foo: %w", err)
+		}
+		defer f.Close()
+
+		image, err := tarball.ImageFromPath(f.Name(), nil)
+		//image, err := daemon.Image(img.imageRef, daemon.WithClient(img.client), daemon.WithUnbufferedOpener())
```
simar7, line 75: What's this?

cmd/fanal/main.go
knqyf263: If we define `func` with nil, it will cause panic.
https://play.golang.org/p/bE51u0pk0Ko

But if we define `func` like the above, it works.
https://play.golang.org/p/uFaa4mhUAoF

Or, we have to do nil check.

```
defer func(){
    if cleanup != nil {
        cleanup()
    }
}
```

cache/cache.go
knqyf263: It makes sure buckets have been created here.
https://github.com/aquasecurity/fanal/blob/26afaa6f737011f7a821e7df4007cda69b5daaa4/cache/cache.go#L60-L64

If it fails to create these buckets, `NewFsCache` is supposed to fail before calling `getLayer`.

extractor/docker/docker.go
knqyf263: If `l` gets to be nil, `LayerByDiffID` will return an error. It means `isCompressed` will not be called.
https://github.com/aquasecurity/fanal/blob/26afaa6f737011f7a821e7df4007cda69b5daaa4/extractor/docker/docker.go#L218-L221

extractor/image/daemon/image_test.go
knqyf263: I've fixed it.
https://github.com/aquasecurity/fanal/pull/96/commits/3558e961a29dd1ab5fae0a14de2c73c89e175a85

extractor/image/daemon/image.go
knqyf263: I'm sorry this is cruft. I removed it https://github.com/aquasecurity/fanal/pull/96/commits/06b40263f9a01c7b25336ffa54798243962f9179.

cache/cache.go
```diff
@@ -75,15 +74,12 @@ func NewFSCache(cacheDir string) (FSCache, error) {
...
+func (fs FSCache) GetLayer(diffID string) (types.LayerInfo, error) {
 	var layerInfo types.LayerInfo
 	err := fs.db.View(func(tx *bolt.Tx) error {
-		// get a decompressed layer ID
-		decompressedBucket := tx.Bucket([]byte(decompressedDigestBucket))
-		layerBucket := tx.Bucket([]byte(layerBucket))
-
 		var err error
-		layerInfo, err = fs.getLayer(decompressedBucket, layerBucket, layerID)
+		layerBucket := tx.Bucket([]byte(layerBucket))
```
simar7, line 81: We should check what `layerBucket` has before we use it. If it's `nil`, `fs.getLayer` will panic.

cache/cache_test.go
```diff
@@ -87,53 +87,6 @@ func TestFSCache_GetLayer(t *testing.T) {
 				},
 			},
 		},
-		{
```
simar7, line 90: We should add some sad paths here. Especially since something like this https://github.com/aquasecurity/fanal/pull/96/files/26afaa6f737011f7a821e7df4007cda69b5daaa4#r405123771 isn't tested and can break.

cache/cache.go
```diff
@@ -111,23 +101,18 @@ func (fs FSCache) getLayer(decompressedBucket, layerBucket *bolt.Bucket, layerID
...
-			decompressedBucket := tx.Bucket([]byte(decompressedDigestBucket))
-			err := decompressedBucket.Put([]byte(layerID), []byte(decompressedLayerID))
-			if err != nil {
-				return xerrors.Errorf("unable to store a pair of compressed/decompressed layer IDs: %w", err)
-			}
-		}
-
 		layerBucket := tx.Bucket([]byte(layerBucket))
-		err = layerBucket.Put([]byte(decompressedLayerID), b)
+		err = layerBucket.Put([]byte(diffID), b)
```
simar7, line 115: ditto, potential for panic.

cmd/fanal/main.go
simar7: I see. I understand.

extractor/docker/docker.go
```diff
@@ -186,41 +180,72 @@ func ApplyLayers(layers []types.LayerInfo) types.ImageDetail {
...
+	conf, err := d.image.ConfigFile()
+	if err != nil {
+		return nil, xerrors.Errorf("unable to get the config file: %w", err)
+	}
+
+	var layerIDs []string
+	for _, d := range conf.RootFS.DiffIDs {
+		layerIDs = append(layerIDs, d.String())
+	}
+	return layerIDs, nil
```
simar7, line 208: These seem to be untested.

extractor/docker/docker.go
simar7: I understand.

extractor/docker/docker_test.go
```diff
@@ -404,76 +406,100 @@ func TestApplyLayers(t *testing.T) {
 }
 
 func TestExtractor_ExtractLayerFiles(t *testing.T) {
+	te, tr := setupDockerEnvironment()
+	defer te.Close()
+	defer tr.Close()
+
+	serverAddr := tr.Listener.Addr().String()
```
simar7, line 413: Just curious, will the httptest server through test{engine,registry} be ready before the tests get run?

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+// To avoid entire loading, this wrapper uses ImageInspectWithRaw and checks image ID and layer IDs.
+type image struct {
+	v1.Image
+	opener  opener
+	inspect types.ImageInspect
+}
+
+type opener func() (v1.Image, error)
+
+func Image(ref name.Reference) (v1.Image, func(), error) {
```
simar7, line 32: We should add some documentation here saying that `cleanup` func must be called by the caller of this function. Otherwise we'll be filling up precious disk space.

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
+package daemon
```
simar7, line 1: This file could use some test coverage.

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+	opener  opener
+	inspect types.ImageInspect
+}
+
+type opener func() (v1.Image, error)
+
+func Image(ref name.Reference) (v1.Image, func(), error) {
+	c, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return nil, nil, xerrors.Errorf("failed to initialize a docker client: %w", err)
```
simar7, line 35: Do you think we should return an empty `cleanup` func instead of `nil`?

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	c.NegotiateAPIVersion(context.Background())
+
+	inspect, _, err := c.ImageInspectWithRaw(context.Background(), ref.Name())
+	if err != nil {
+		return nil, nil, err
```
simar7, line 42: ditto

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	f, err := ioutil.TempFile("", "fanal-*")
+	if err != nil {
+		return nil, nil, err
```
simar7, line 47: ditto

extractor/image/daemon/image.go
simar7, line 35:
```suggestion
		return nil, func(){}, xerrors.Errorf("failed to initialize a docker client: %w", err)
```

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func (img *image) populateImage() (err error) {
+	mu.Lock()
+	defer mu.Unlock()
+
+	if img.Image != nil {
+		return nil
+	}
```
simar7, line 89: What does this do?

extractor/image/daemon/image.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"golang.org/x/xerrors"
+)
+
+var mu sync.Mutex
```
simar7, line 18: I'm curious why do we need a mutex? What call sites does this package have?

cache/cache.go
knqyf263: Is your question duplicated? Or, do you mean we should check it in any case?
https://github.com/aquasecurity/fanal/pull/96#discussion_r404431270

cache/cache.go
knqyf263: https://github.com/aquasecurity/fanal/pull/96#discussion_r405443752

cache/cache_test.go
knqyf263: I added https://github.com/aquasecurity/fanal/pull/96/commits/77f4a81848d5c12820bbd362131df33526962a64

extractor/docker/docker.go
knqyf263: They are tested through `analyzer_test.go`.

extractor/docker/docker_test.go
knqyf263: I think httptest.NewServer waits until it can accept a request.
https://github.com/golang/go/blob/master/src/net/http/httptest/server.go#L304

extractor/image/daemon/image.go
knqyf263: I added. https://github.com/aquasecurity/fanal/pull/96/commits/8bdb3a5920606dfb68ae433f859092a3b1bb21a0

extractor/image/daemon/image.go
knqyf263: Ah, you're right. It's better than nil.
https://github.com/aquasecurity/fanal/pull/96/commits/f3f7adc014ee85385d8fb0e8da047bb040e6c8fe

extractor/image/daemon/image.go
knqyf263: https://github.com/aquasecurity/fanal/pull/96/commits/f3f7adc014ee85385d8fb0e8da047bb040e6c8fe

extractor/image/daemon/image.go
knqyf263: https://github.com/aquasecurity/fanal/pull/96/commits/f3f7adc014ee85385d8fb0e8da047bb040e6c8fe

extractor/image/daemon/image.go
knqyf263: `LayerByDiffID` is called by some goroutines at the same time to reduce the download time. If we don't have `sync.Mutex`, `populateImage` will be called many times. It means the same image is written down in a local file system many times.

extractor/image/daemon/image.go
knqyf263: Initializing `img.Image` is a high cost, so we should avoid it as possible as we can. If `img.Image` is not nil, we don't have to initialize it again.
https://github.com/aquasecurity/fanal/pull/96/commits/fdc17a2c8a87d22f309a1f7fbe22c16e4689d303

extractor/image/daemon/image.go
simar7: Got it.

extractor/docker/docker_test.go
simar7: I meant to say if the server is not ready yet and the request is made prior to it being ready, the request would fail. You can see an example here https://github.com/aquasecurity/testdocker/blob/master/registry/server_test.go#L51-L58

cache/cache.go
simar7: I think it's OK. It is a potential for panic but since we already make sure NewFsCache creates it, as long as we call NewFsCache before using FsCache{} it should be fine.

extractor/image/daemon/image.go
simar7: I got it. Could you add a little comment for this?

simar7:
Looks like the benchmarks aren't happy
image

extractor/docker/docker_test.go
knqyf263: What do you mean? Your example uses goroutine and it is the cause. But I meant `httptest` waited until the server is ready unless it is called in goroutine. It means the next line will not be called and a request will not be sent until the server can accept a request.

I think this example will never fail.
https://play.golang.org/p/QDDTLsmMT2h

cache/cache.go
knqyf263: OK. If we face a panic, it is the time when we should handle it.

extractor/image/daemon/image.go
knqyf263: Roger. I added https://github.com/aquasecurity/fanal/pull/96/commits/3833f9cf20d4a262151440f267761599b7a7019d

extractor/docker/docker_test.go
simar7: I got it.

PR #98, opened 2020-04-01
fix(bench): measure initializing structs

knqyf263:
## Background
Currently, our benchmark measures only `Analyze()` function, but this PR also measures the initialize cost of each struct. We're working on replacing containers/image with google/go-containerregistry now. I found `containers/image` did most of the processes in the constructor, while google/go-containerregistry did them after the constructor. It means containers/image looks faster than google/go-containerregistry in our benchmark.

Also, this PR makes the benchmark look worse because it includes the initial cost now. So, I skip the test once.

## TODO
- [x] remove some cases to reduce the test time
- [x] measure initializing structs
- [x] force benchmark tests to pass

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -22,4 +22,4 @@ jobs:
       run: curl -sfL https://raw.githubusercontent.com/knqyf263/cob/master/install.sh | sudo sh -s -- -b /usr/local/bin
 
     - name: Run Benchmark
-      run: cob --threshold 0.7 --base origin/master --bench-cmd make --bench-args test-performance
+      run: cob --threshold 0.7 --base origin/master --bench-cmd make --bench-args test-performance || exit 0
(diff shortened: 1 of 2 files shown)


.github/workflows/bench.yml
```diff
@@ -22,4 +22,4 @@ jobs:
       run: curl -sfL https://raw.githubusercontent.com/knqyf263/cob/master/install.sh | sudo sh -s -- -b /usr/local/bin
 
     - name: Run Benchmark
-      run: cob --threshold 0.7 --base origin/master --bench-cmd make --bench-args test-performance
+      run: cob --threshold 0.7 --base origin/master --bench-cmd make --bench-args test-performance || exit 0
```
simar7, line 25: This is temporary right? You will remove before the PR is merged?

knqyf263: Yes, it's temporary, but I will remove it in the next PR because the test in this PR must fail. It is obvious and intended, but it looks red 🔴. So, I suppress the error once.
image

PR #100, opened 2020-04-07
chore(actions): add macOS

knqyf263:


knqyf263:
I found macOS runner in GitHub Actions didn't support Docker for Mac.
https://github.community/t5/GitHub-Actions/Is-it-possible-to-install-and-configure-Docker-on-MacOS-runner/td-p/39002

PR #101, opened 2020-04-20
feat(image): support OCI Image Format

knqyf263:
You can test this PR with the following commands.

```
$ skopeo copy docker-daemon:alpine:3.11 oci:/tmp/alpine.oci
$ go run cmd/fanal/main.go -f /tmp/alpine.oci
&{Family:alpine Name:3.11.5}
via image Packages: 18
```

Issue: https://github.com/aquasecurity/trivy/issues/465


extractor/image/image.go
```diff
@@ -93,3 +109,33 @@ func fileOpener(fileName string) func() (io.ReadCloser, error) {
...
+	if err != nil {
+		return nil, xerrors.Errorf("invalid index.json: %w", err)
+	}
+
+	if len(m.Manifests) == 0 {
+		return nil, xerrors.New("no valid manifest")
+	}
+
+	// Support only first image
+	h := m.Manifests[0].Digest
```
simar7, line 134: If we only support first image, does this still make us say we're OCI compliant?

extractor/image/image_test.go
```diff
@@ -118,6 +118,12 @@ func TestNewDockerArchiveImage(t *testing.T) {
 				fileName: "../testdata/alpine-310.tar.gz",
 			},
 		},
+		{
+			name: "happy path with OCI Image Format",
+			args: args{
+				fileName: "../testdata/test.oci",
+			},
+		},
```
simar7, line 126: Could you also add some sad paths?

extractor/image/image.go
```diff
@@ -65,7 +66,22 @@ func NewDockerImage(ctx context.Context, imageName string, option types.DockerOp
 	return img, func() {}, nil
 }
 
+// TODO: rename NewDockerArchiveImage to NewArchiveImage
 func NewDockerArchiveImage(fileName string) (v1.Image, error) {
```
simar7, line 70: `NewArchiveImage` sounds good to me.

```suggestion
func NewArchiveImage(fileName string) (v1.Image, error) {
```

extractor/image/image_test.go
knqyf263: We already have a sad path. Isn't it enough?
https://github.com/aquasecurity/fanal/blob/387cf0c049fd5e9073ba760125964f0f3b81bfc9/extractor/image/image_test.go#L127-L133

extractor/image/image.go
knqyf263: `containers/image` does the same thing if any tag/digests are not specified, but they also support multiple images.
https://github.com/containers/image/blob/d13a90f787a7d3cd58e6f13c7cd1f2307b74a1ef/oci/layout/oci_transport.go#L188-L194

If we support multiple images, it makes the implementation a bit complicated. A user needs to specify which image is scanned by digest like the following.

```
$ fanal -f /tmp/alpine.oci@sha256:24fbbcafb45a7030f7d53035753733950ec23320cecb4d10d0f1d2f1da9c62f0
```

This is because a tag like `latest` is optional for OCI, so we need a digest. IMO, there is only 1 image in OCI directory in many cases.

```
// select the first image
$ fanal -f /tmp/alpine.oci
```

I feel it is enough at the moment. If someone raises an issue, we can support multiple images.

extractor/image/image.go
knqyf263: Done https://github.com/aquasecurity/fanal/pull/101/commits/d9baf353482f82019971d255076eeeb63af2d5ec

extractor/image/image.go
simar7: OK! I was just curious.

extractor/image/image_test.go
simar7: I feel it's not. There are still other ways the function can fail. See the coverage as shown below. Red indicates no coverage.
image

knqyf263: Do you think we should go through all paths? Should I add 4 more cases?

simar7: I tried to add tests but I noticed another issue. Currently all errors are masked upstream here: https://github.com/aquasecurity/fanal/pull/101/files#diff-525dd35201b72b43ad63b85dd2720ba6R75-R80 into a single error of `"invalid archive"`.

So we will never be able to see the underlying error that happened in `tryOCI` func. This would be an example to test our return types in such a case. 

I agree that the library is well tested but it doesn't guard against regressions that happens because of future refactorings of `tryOCI` func or in this case where the error isn't what we expected it to be.

extractor/image/image.go
simar7: Can you also rename the test accordingly? https://github.com/aquasecurity/fanal/pull/101/files#diff-525dd35201b72b43ad63b85dd2720ba6L105

`func TestNewDockerArchiveImage(t *testing.T) {`
to 
`func TestNewArchiveImage(t *testing.T) {`

knqyf263: Thanks. I've fixed it.
https://github.com/aquasecurity/fanal/pull/101/commits/9163201742aada50207d588bc5d2d5366270140a

extractor/image/image_test.go
knqyf263: I didn't have a way to handle multiple errors, so I introduced hashicorp/go-multierror.
https://github.com/aquasecurity/fanal/pull/101/commits/12c9696ce44781570891ea24e76d2146159e9ebb

If you feel we should add more tests, please point it out. Of course, you can add it yourself.

simar7: Cool! I added a few more tests: https://github.com/aquasecurity/fanal/pull/101/commits/98fd6e4bb85627e05c733560cfd684b8fa650003

extractor/image/image.go
```diff
@@ -93,3 +119,33 @@ func fileOpener(fileName string) func() (io.ReadCloser, error) {
...
+func tryOCI(fileName string) (v1.Image, error) {
+	lp, err := layout.FromPath(fileName)
+	if err != nil {
+		return nil, xerrors.Errorf("unable to open %s as an OCI Image: %w", fileName, err)
+	}
+
+	index, err := lp.ImageIndex()
+	if err != nil {
+		return nil, xerrors.Errorf("unable to retrieve index.json: %w", err)
+	}
```
simar7, line 132: I'm curious as to when would this error happen? I tried to test it by removing `index.json` from a test image but that case is actually handled by the previous if condition.

knqyf263: I'm not sure the case is useful, but I added it. https://github.com/aquasecurity/fanal/pull/101/commits/0e7df0c642f570e1930d83fe32ef7d40477b93a7

An index.json with no permission should also fail at `ImageIndex()`.

PR #102, opened 2020-04-22
feat(image): support registry token

knqyf263:
If `RegistryToken` is passed directly, it will be used to authenticate with a docker registry.

Blocker: https://github.com/aquasecurity/testdocker/pull/2

Issue: https://github.com/aquasecurity/trivy/issues/453


extractor/image/image_test.go
```diff
@@ -102,6 +109,96 @@ func TestNewDockerImage(t *testing.T) {
...
+
+	type args struct {
+		imageName string
+		option    types.DockerOption
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    v1.Image
+		wantErr bool
```
danielpacak, line 149: NIT: I'm usually asserting on the error message. This helps in making sure we have appropriate & meaningful error messages when troubleshooting

knqyf263: Thank you for the feedback. I added the assert. I usually assert error messages, but I was slacked off a little 😄

PR #109, opened 2020-05-24
feat: support local filesystem and remote git repository [PART 1]

knqyf263:
## Overview
Add `tar` and `filesystem` walker.

## Note
This is a stacked pull request.

## Parent

## TODO
- [x] fs walker
- [x] tar walker


walker/tar.go
```diff
@@ -0,0 +1,79 @@
...
+			if err != nil {
+				return nil, nil, xerrors.Errorf("failed to analyze file: %w", err)
+			}
+		}
+	}
+
+	return opqDirs, whFiles, nil
+}
+
+// fileOnceOpener reads a file once and the content is shared so that some analyzers can use the same data
```
simar7, line 64:
```suggestion
// tarOnceOpener reads a file once and the content is shared so that some analyzers can use the same data
```

walker/tar.go
```diff
@@ -0,0 +1,79 @@
...
+func tarOnceOpener(r io.Reader) func() ([]byte, error) {
+	var once sync.Once
+	var b []byte
+	var err error
+
+	return func() ([]byte, error) {
+		once.Do(func() {
```
simar7, line 71: I'm not familiar with `once.Do()` so for my understanding: Do subsequent calls after the first one fail? How is data shared amongst callers?

walker/tar.go
knqyf263: Thanks. I'll apply this change after merging #110 - #114 to avoid conflict.

walker/tar.go
knqyf263: It's closure and `once`, `b` and `err` are shared among callers. If `once.Do` fails, `err` is not nil, so it always returns the error. But if this function returns an error, the program exits anyway.

PR #110, opened 2020-05-24
feat: support local filesystem and remote git repository [PART 2]

knqyf263:
## Overview
Merge `OSAnalyzer`, `PkgAnalyze` and `LibraryAnalyzer` into `analyzer` interface.

```
type analyzer interface {
	Name() string
	Analyze(content []byte) (AnalyzeReturn, error)
	Required(filePath string, info os.FileInfo) bool
}
```

## Note
- The tests which are commented out are fixed in https://github.com/aquasecurity/fanal/pull/113.

## Parent


analyzer/analyzer.go
```diff
@@ -30,310 +23,131 @@ var (
...
 }
 
-func GetPackages(filesMap extractor.FileMap) ([]types.PackageInfo, error) {
-	var results []types.PackageInfo
-	for _, analyzer := range pkgAnalyzers {
-		pkgMap, err := analyzer.Analyze(filesMap)
-
-		// Differentiate between a package manager not being found and another error
-		if err != nil && err == ErrNoPkgsDetected {
+func AnalyzeConfig(targetOS types.OS, configBlob []byte) []types.Package {
```
simar7, line 136: Could you remind me again what is this used for?

simar7: On a sidenote, my IDE says only Alpine currently implements `analyzer.Analyze` what about other OSes?

analyzer/library/bundler/bundler.go
```diff
@@ -2,48 +2,41 @@ package bundler
...
+	return analyzer.AnalyzeReturn{
+		Libraries: libs,
+	}, nil
 }
 
-func (a bundlerLibraryAnalyzer) RequiredFiles() []string {
-	return []string{"Gemfile.lock"}
+func (a bundlerLibraryAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	fileName := filepath.Base(filePath)
+	return utils.StringInSlice(fileName, requiredFiles)
```
simar7, line 39: A general observation between the various library/*.go files is that there's a lot of duplicate logic especially for functions like `Analyze`, `Required`. Could we DRY this up somehow? If we end up doing so it will also become easier to write a tests (less duplicate setups to keep).

analyzer/analyzer.go
```diff
@@ -1,26 +1,19 @@
...
 )
 
 var (
-	osAnalyzers      []OSAnalyzer
-	pkgAnalyzers     []PkgAnalyzer
-	libAnalyzers     []LibraryAnalyzer
-	commandAnalyzers []CommandAnalyzer
-	additionalFiles  []string
+	analyzers       []analyzer
+	configAnalyzers []configAnalyzer
```
simar7, line 16: Design: Since we are modifying this I would like to propose that we should eliminate global package state if we can. I think we can achieve this by exposing a function that can take in a argument of `[]analyzer` and the user of the library can pass in the various analyzers they would like to register.

More commentary here: https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables

analyzer/analyzer.go
```diff
@@ -30,310 +23,131 @@ var (
...
 		}
-		layers = append(layers, layer)
-	}
-
-	mergedLayer := docker.ApplyLayers(layers)
-	if mergedLayer.OS == nil {
-		return mergedLayer, ErrUnknownOS // send back package and apps info regardless
-	} else if mergedLayer.Packages == nil {
-		return mergedLayer, ErrNoPkgsDetected // send back package and apps info regardless
-	}
```
simar7, line 254: What happens in this use case now?

test/integration/benchmark_test.go
```diff
@@ -25,10 +24,11 @@ import (
 	_ "github.com/aquasecurity/fanal/analyzer/library/yarn"
 	_ "github.com/aquasecurity/fanal/analyzer/os/alpine"
 	_ "github.com/aquasecurity/fanal/analyzer/os/amazonlinux"
-	_ "github.com/aquasecurity/fanal/analyzer/os/debianbase"
+	_ "github.com/aquasecurity/fanal/analyzer/os/debian"
```
simar7, line 27: Have you renamed `debianbase` to `debian`?

analyzer/os/photon/photon.go
```diff
@@ -3,53 +3,57 @@ package photon
...
-			}
-
-			if photonName != "" && strings.HasPrefix(line, "VERSION_ID=") {
-				return types.OS{
+
+		if photonName != "" && strings.HasPrefix(line, "VERSION_ID=") {
+			return analyzer.AnalyzeReturn{
+				OS: types.OS{
 					Family: photonName,
 					Name:   strings.TrimSpace(line[11:]),
```
simar7, line 45: What happens if `line` doesn't have enough length to do [11:]?

analyzer/os/amazonlinux/amazonlinux.go
```diff
@@ -4,51 +4,63 @@ import (
...
+		// Only Amazon Linux Prefix
+		if strings.HasPrefix(line, "Amazon Linux release 2") {
+			return types.OS{
+				Family: aos.Amazon,
+				Name:   fmt.Sprintf("%s %s", fields[3], fields[4]),
+			}, nil
+		} else if strings.HasPrefix(line, "Amazon Linux") {
+			return types.OS{
+				Family: aos.Amazon,
+				Name:   strings.Join(fields[2:], " "),
```
simar7, line 53: We should check what size `fields` is before we operate on individual slice items.

analyzer/analyzer.go
simar7: Oh I guess this is reintroduced here? https://github.com/aquasecurity/fanal/pull/111/files?file-filters%5B%5D=.go#diff-8d7f91425a215d1c8e3d51d7b49ebfa2R28-R33 

Am I correct?

analyzer/analyzer.go
```diff
@@ -30,310 +23,131 @@ var (
...
-		os, err := analyzer.Analyze(filesMap)
+		ret, err := analyzer.Analyze(b)
 		if err != nil {
 			continue
 		}
-		return &os
+		result.Merge(ret.ConvertToResult(analyzer.Name(), filePath))
 	}
-	return nil
+	return result, nil
```
simar7, line 133: This will create a copy of the result which includes a `sync.Mutex` which will be ended up getting copied. I think we should return a reference here instead to avoid that situation. Or we could make `sync.Mutex` a reference within the `AnalysisResult` struct.

analyzer/analyzer.go
knqyf263: `Analyzer` analyzers files in a container image such as `/var/lib/dpkg/status`, but an image has metadata in `config.json`. This is managed separately from the layer content and includes history such as `RUN`, `COPY` and `ENV` used in Dockerfile. `AnalyzeConfig` is intended to analyze such a history.

Especially, the target is the following case.

```
RUN apk add openssl && \
        wget https://example.com/anysoftware.tar.gz && \
        tar xvf anysoftware.tar.gz && cd anysoftware && \
       ./configure --with-ssl && make && make install && \
       apk del openssl
```

In this case, `openssl` is linked to `anysoftware` statically, but it doesn't exist in the merged layer since this package is deleted. So, `AnalyzeConfig` parses `RUN` lines and extracts deleted packages.

>On a sidenote, my IDE says only Alpine currently implements analyzer.Analyze what about other OSes?
Right. As you can see, `RUN` doesn't show which version is installed. But Alpine stores only the latest version in their repository. So, we can know the version from `created_at`. I think it is difficult for other OSes.

analyzer/analyzer.go
knqyf263: Right. `ApplyLayers` was in `analyzer.go`, but I moved it in the dedicated file, `applier.go`.

test/integration/benchmark_test.go
knqyf263: I split `debianbase` into `debian` and `ubuntu`.

analyzer/library/bundler/bundler.go
knqyf263: I thought it was flexible to implement in each library/*.go, but I'll try DRY.

analyzer/analyzer.go
knqyf263: Yes, I usually avoid package scoped variables, but this is `Registration`pattern and many go standard packages use it.
https://sourcegraph.com/github.com/golang/go/-/blob/src/database/sql/sql.go#L35

The blog author doesn't like the `Registration` pattern, but I feel there are some advantages. It makes introducing plugin easy. A build tag can toggle whether we use the analyzer or not.

```
// +build commercial

package myanalyzer

func init() {
    analyzer.Register(myAnalyzer{})
}

...
```

```
$ go build // myanalyzer is not used
$ go build -tags commercial  // myanalyzer is used
```

Also, Go has `plugin` package.
https://golang.org/pkg/plugin/

```
// +build commercial

package myanalyzer

func init() {
    analyzer.Register(myAnalyzer{})
}

...
```

```
$ go build -buildmode=plugin -o myanalyzer.so myanalyzer.go
```

```
func main() {
    files, err := ioutil.ReadDir("plugin/")
    if err != nil {
        panic(err)
    }

    for _, file := range files {
        // init() is called
        p, err := plugin.Open(file.Name())
        if err != nil {
            fmt.Fprintln(os.Stderr, err)
            os.Exit(-1)
        }
    }
}
...
```

```
$ go run main.go // myanalyze is not used
$ mv myanalyzer.so plugin/
$ go run main.go // myanalyzer is used
```

I think Go maintainers also know a package scoped variable is not good, but they selected extensibility.

analyzer/os/photon/photon.go
knqyf263: Just empty.
https://play.golang.org/p/UclQNFPuCR_O

analyzer/os/amazonlinux/amazonlinux.go
knqyf263: I've fixed it in https://github.com/aquasecurity/fanal/pull/110/commits/d1bdfd8fbc15a7c2333f86502e86161597fcee25
https://play.golang.org/p/K2DC8_7ajCW

analyzer/library/bundler/bundler.go
knqyf263: I tried the following, but I feel like it doesn't help so much. The original function is small enough, so we can't reduce so many lines. Do you have any idea?

```
func analyze(content []byte, parse func(r io.Reader) ([]godeptypes.Library, error)) (analyzer.AnalyzeReturn, error) {
	r := bytes.NewBuffer(content)
	libs, err := parse(r)
	if err != nil {
		return analyzer.AnalyzeReturn{}, xerrors.Errorf("error with a lock file: %w", err)
	}
	return analyzer.AnalyzeReturn{
		Libraries: libs,
	}, nil
}

type bundlerLibraryAnalyzer struct{}

func (a bundlerLibraryAnalyzer) Analyze(content []byte) (analyzer.AnalyzeReturn, error) {
	ret, err := analyze(content, bundler.Parse)
	if err != nil {
		return analyzer.AnalyzeReturn{}, xerrors.Errorf("unable to parse Gemfile.lock: %w", err)
	}
	return ret, nil
}
```

analyzer/analyzer.go
knqyf263: Thanks. I fixed it.
https://github.com/aquasecurity/fanal/pull/110/commits/ae13a5383a54b7c69ab1c004b3e9955a96596f5f

analyzer/library/bundler/bundler.go
simar7: I like this idea. I pushed some changes here https://github.com/aquasecurity/fanal/pull/110/commits/0f897b6e4727fe9fd216f2a050f6d4a3455a9d0b

PR #111, opened 2020-05-24
feat: support local filesystem and remote git repository [PART 3]

knqyf263:
## Overview
Add `Applier` to apply layers.

## Note
- The tests which are commented out are fixed in https://github.com/aquasecurity/fanal/pull/113.

## Parent


applier/docker.go
```diff
@@ -0,0 +1,122 @@
+package applier
```
simar7, line 1: It's nice to see we can do major refactors and re-writes like this without breaking functionality since `docker_test.go` exists! 🤞

PR #112, opened 2020-05-24
feat: support local filesystem and remote git repository [PART 4]

knqyf263:
## Overview
Add `image`, `local` and `remote` artifacts

## Note

## Parent


artifact/local/fs.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	// get hostname
+	var hostName string
+	b, err := ioutil.ReadFile(filepath.Join(a.dir, "etc", "hostname"))
+	if err != nil {
+		hostName = a.dir
+	} else {
+		hostName = strings.TrimSpace(string(b))
+	}
```
simar7, line 82: What's this for?

artifact/remote/testdata/test.git/hooks/applypatch-msg.sample
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
simar7, line 1: Do we really need all hooks in the `test.git` repo? I was able to run tests just fine without them.

artifact/local/fs.go
knqyf263: if you scan filesystem inside a container like `trivy filesystem /`, it shows `/ (alpine:3.11)`. I thought we should display the hostname. e.g. `web01 (alpine:3.11)`.

artifact/remote/testdata/test.git/hooks/applypatch-msg.sample
knqyf263: Ah, it is unnecessary. I've just run `git init`. I'll remove them.

knqyf263: I removed them https://github.com/aquasecurity/fanal/pull/112/commits/1dd8e2e51167244d38e40861f3dbc7fd191c1196, but the test still passes. Thanks!

artifact/local/fs.go
simar7: I see. How about using `os.Hostname` in that case?

knqyf263: I didn't know it. I'll use it. Thanks!

knqyf263: Oh, I'm sorry, but I found `os.Hostname` doesn't work in our case. When a user scans `/app/python_project`, we shouldn't show the hostname of the local machine. They focus on scanning the project and don't want to know the hostname.

Also, they may export a container's filesystem and scan the directory. In that case, we should display the hostname of the container, not the hostname of the machine running the container.

For example,
```
$ hostname
knqyf263's Mac
$ docker export $(docker create alpine:3.7) |tar -C /tmp/rootfs -xvf -
$ ls /tmp/rootfs
bin/  dev/  etc/  home/  lib/  media/  mnt/  proc/  root/  run/  sbin/  srv/  sys/  tmp/  usr/  var/
$ trivy fs /tmp/rootfs
```

In this case, we shouldn't display `knqyf263's Mac`, but the content of `/tmp/rootfs/etc/hostname`. If we use `os.Hostname()`, it returns `knqyf263's Mac`.

PR #119, opened 2020-06-09
fix: Due read after write consistency in S3 missingLayers called the …

oranmoshai:
…actual object that created cache 403 response

This change creating index file for each object so missingLayers will not hit object that not exist.

### DIFF
--- a/cache/s3.go
+++ b/cache/s3.go
@@ -129,7 +129,7 @@ func (c S3Cache) MissingBlobs(artifactID string, blobIDs []string) (bool, []stri
 		}
 		blobInfo, err := c.GetBlob(blobID)
 		if err != nil {
-			return true, missingBlobIDs, xerrors.Errorf("failed to get blob from the cache: %w", err)
+			return true, missingBlobIDs, xerrors.Errorf("the blob object (%s) doesn't exist in S3 even though the index file exists: %w", blobID, err)
 		}
 		if blobInfo.SchemaVersion != types.BlobJSONSchemaVersion {
 			missingBlobIDs = append(missingBlobIDs, blobID)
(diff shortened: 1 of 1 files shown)


cache/s3.go
```diff
@@ -109,34 +106,48 @@ func (c S3Cache) GetArtifact(artifactID string) (types.ArtifactInfo, error) {
...
-		blobInfo, err := c.GetBlob(blobID)
+		err := c.getIndex(blobID, blobBucket)
 		if err != nil {
 			// error means cache missed blob info
 			missingBlobIDs = append(missingBlobIDs, blobID)
 			continue
 		}
+		blobInfo, err := c.GetBlob(blobID)
+		if err != nil {
+			return true, missingBlobIDs, xerrors.Errorf("failed to get blob from the cache: %w", err)
```
knqyf263, line 132: You want to return an error instead of `continue` because this should never happen, right?

```suggestion
			return true, missingBlobIDs, xerrors.Errorf("the blob object (%s) doesn't exist in S3 even though the index file exists: %w", blobID, err)
```

cache/s3.go
```diff
@@ -109,34 +106,48 @@ func (c S3Cache) GetArtifact(artifactID string) (types.ArtifactInfo, error) {
...
-	artifactInfo, err := c.GetArtifact(artifactID)
+	err := c.getIndex(artifactID, artifactBucket)
+	// error means cache missed artifact info
 	if err != nil {
-		// error means cache missed artifact info
 		return true, missingBlobIDs, nil
 	}
+	artifactInfo, err := c.GetArtifact(artifactID)
+	if err != nil {
+		return true, missingBlobIDs, xerrors.Errorf("failed to get artifact from the cache: %w", err)
```
knqyf263, line 146:
```suggestion
		return true, missingBlobIDs, xerrors.Errorf("the artifact object (%s) doesn't exist in S3 even though the index file exists: %w", artifactID, err)
```

Issue #126, opened 2020-07-09
Handle removed packages in Debian/Ubuntu


knqyf263:
In Debian, the package remains in `/var/lib/dpkg/status` even after the package is removed without `--purge`

```
$ docker run --rm -it debian:buster bash
root@2209f993228f:/# apt-get update
root@2209f993228f:/# apt-get install git
root@2209f993228f:/# apt-get remove git
root@2209f993228f:/# dpkg -l | grep git
rc  git                       1:2.20.1-2+deb10u3          amd64        fast, scalable, distributed revision control system
```

`fanal` assumes that the package is installed if it is listed in `/var/lib/dpkg/status`. We need to handle `Status` line.

```
root@2209f993228f:/# cat /var/lib/dpkg/status | grep -A3 -e "Package: git$"
Package: git
Status: deinstall ok config-files
Priority: optional
Section: vcs
```

If the package is installed, `Status` should be "installed".

```
root@2209f993228f:/# cat /var/lib/dpkg/status | grep -A3 -e "Package: bash$"
Package: bash
Essential: yes
Status: install ok installed
Priority: required
```

knqyf263:
Hi @rahul2393, could you work on it? At first, you have to create the image including a removed package.

```
$ cat Dockerfile
FROM debian:buster
RUN apt-get -y update && apt-get install git
RUN apt-get remove git
$ docker build -t issue-126 .
```

Then, you scan the image by fanal. But fanal just shows the number of installed packages as is. You have to edit `cmd/fanal/main.go` so that it can display the names of installed packages.

```
$ go run cmd/fanal/main.go image issue-126
```

After you make sure `git` is displayed by fanal, you can fix this problem.

PR #127, opened 2020-07-14
Check status for dpkg package

rahul2393:
[Issue](https://github.com/aquasecurity/fanal/issues/126)

### DIFF
--- a/analyzer/pkg/dpkg/dpkg.go
+++ b/analyzer/pkg/dpkg/dpkg.go
@@ -107,9 +107,13 @@ func (a debianPkgAnalyzer) parseDpkgPkg(scanner *bufio.Scanner) (pkg *types.Pack
 		} else if strings.HasPrefix(line, "Version: ") {
 			version = strings.TrimPrefix(line, "Version: ")
 		} else if strings.HasPrefix(line, "Status: ") {
-			isInstalled = strings.HasPrefix(strings.TrimPrefix(line, "Status: "), "install")
+			for _, ss := range strings.Fields(strings.TrimPrefix(line, "Status: ")) {
+				if ss == "deinstall" || ss == "purge" {
+					isInstalled = false
+					break
+				}
+			}
 		}
-
 		if !scanner.Scan() {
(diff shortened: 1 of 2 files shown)


analyzer/pkg/dpkg/testdata/dpkg
```diff
@@ -2130,4 +2130,26 @@ Description: Micro string library: shared library
  This package contains the shared library for ustr.
 Homepage: http://www.and.org/ustr/
 
+Package: git
+Status: uninstall ok installed
```
knqyf263, line 2134: I prefer using an actual value. Did you see this value in `Status`? AFAIK, it's like `Status: deinstall ok config-files`.

analyzer/pkg/dpkg/dpkg.go
```diff
@@ -105,14 +106,16 @@ func (a debianPkgAnalyzer) parseDpkgPkg(scanner *bufio.Scanner) (pkg *types.Pack
 			}
 		} else if strings.HasPrefix(line, "Version: ") {
 			version = strings.TrimPrefix(line, "Version: ")
+		} else if strings.HasPrefix(line, "Status: ") {
+			isInstalled = strings.HasPrefix(strings.TrimPrefix(line, "Status: "), "install")
```
knqyf263, line 110: Is the order of statuses always the same? I'm worried that the package is not detected when `Status` is like `ok install installed` In that case, we cannot detect any vulnerability of the package. The best way is we'll check the specification. If we can't find it, we should check with Debian 8, 9, 10 and Ubuntu 16.04, 18.04, 20.04.

analyzer/pkg/dpkg/testdata/dpkg
rahul2393: @knqyf263 fixed

analyzer/pkg/dpkg/dpkg.go
rahul2393: @knqyf263 Updated

PR #128, opened 2020-07-18
Added skip_dir in image artifacts scan

rahul2393:
[issue](https://github.com/aquasecurity/trivy/issues/244)


artifact/image/image.go
```diff
@@ -19,14 +19,16 @@ import (
...
 type Artifact struct {
-	image image.Image
-	cache cache.ArtifactCache
+	image           image.Image
+	cache           cache.ArtifactCache
+	skipDirectories []string
 }
 
-func NewArtifact(img image.Image, c cache.ArtifactCache) artifact.Artifact {
+func NewArtifact(img image.Image, c cache.ArtifactCache, skipDirectories ...string) artifact.Artifact {
```
knqyf263, line 27:
```suggestion
func NewArtifact(img image.Image, c cache.ArtifactCache, skipDirectories []string) artifact.Artifact {
```

This parameter will be passed from a CLI option in Trivy. In that case,  `skipDirectories` should be already []string. I think we don't have to use a variadic function.

walker/tar.go
```diff
@@ -16,7 +16,7 @@ const (
 	wh  string = ".wh."
 )
 
-func WalkLayerTar(layer io.Reader, analyzeFn WalkFunc) ([]string, []string, error) {
+func WalkLayerTar(layer io.Reader, analyzeFn WalkFunc, skipDirectories ...string) ([]string, []string, error) {
```
knqyf263, line 19: nit: just my preference, but I prefer a function to be at the end for readability.
```suggestion
func WalkLayerTar(layer io.Reader, skipDirectories []string, analyzeFn WalkFunc) ([]string, []string, error) {
```

walker/walk.go
```diff
@@ -26,6 +26,11 @@ func isIgnored(filePath string) bool {
 		if utils.StringInSlice(path, ignoreDirs) {
 			return true
 		}
+		if skipDirectories != nil {
+			if utils.StringInSlice(path, skipDirectories) {
+				return true
+			}
+		}
```
knqyf263, line 33: Even if `skipDirectories` is nil, `StringInSlice` just returns false.
https://play.golang.org/p/ujs5JjC8Vul

So, it looks like we don't need the nil check.

```suggestion
		if utils.StringInSlice(path, skipDirectories) {
				return true
		}
```

artifact/image/image.go
```diff
@@ -19,14 +19,16 @@ import (
 )
 
 type Artifact struct {
-	image image.Image
-	cache cache.ArtifactCache
+	image           image.Image
+	cache           cache.ArtifactCache
+	skipDirectories []string
```
knqyf263, line 24: I considered it again and now I feel `skipDirectories` is more `Inspect` stuff. What if we pass it as the argument of `Inspect`? We'll likely have more options for `Inspect`, so it might be better to provide `InspectOption`.
However, I haven't looked into what change we need. Let me know if we need any bigger change than the current implementation.

walker/walk_test.go
```diff
@@ -10,18 +10,22 @@ import (
...
-		assert.True(t, isIgnored(fp))
+		assert.True(t, isIgnored(fp, nil))
 	}
 
 	for _, fp := range []string{"foo", "foo/bar"} {
-		assert.False(t, isIgnored(fp))
+		assert.False(t, isIgnored(fp, nil))
+	}
+
+	for _, fp := range []string{"foo", "foo/bar"} {
```
knqyf263, line 28: How about `/foo/bar`? This value is passed from the CLI option. I think a user usually adds the prefix `/`, but we can't predict it. We should be able to handle both.

PR #129, opened 2020-08-12
Revert "Added skip_dir in image artifacts scan"

knqyf263:
Reverts aquasecurity/fanal#128 in favor of https://github.com/aquasecurity/trivy/pull/595

This implementation is affected by the cache. For example,

```
$ go build -o fanal cmd/fanal/main.go 
$ ./fanal image fluent/fluentd:edge
fluent/fluentd:edge
&{Family:alpine Name:3.9.6}
via image Packages: 31
bundler (usr/lib/ruby/gems/2.5.0/gems/async-http-0.50.7/examples/fetch/Gemfile.lock): 26
bundler (usr/lib/ruby/gems/2.5.0/gems/http_parser.rb-0.6.0/Gemfile.lock): 14
$ ./fanal --skip-directories "/usr/lib/ruby/gems" image fluent/fluentd:edge
fluent/fluentd:edge
&{Family:alpine Name:3.9.6}
via image Packages: 31
bundler (usr/lib/ruby/gems/2.5.0/gems/async-http-0.50.7/examples/fetch/Gemfile.lock): 26
bundler (usr/lib/ruby/gems/2.5.0/gems/http_parser.rb-0.6.0/Gemfile.lock): 14
```

The command with `--skip-directories` should not display lock files, but it does due to the cache.

Also, it affects the cache.

```
$ ./fanal -clear image
$ ./fanal --skip-directories "/usr/lib/ruby/gems" image fluent/fluentd:edge
fluent/fluentd:edge
&{Family:alpine Name:3.9.6}
via image Packages: 31
$ ./fanal image fluent/fluentd:edge
fluent/fluentd:edge
&{Family:alpine Name:3.9.6}
via image Packages: 31
```

The command without `--skip-directories` should display lock files, but it doesn't.

I noticed that we should not skip directories during analysis process. If we want to suppress vulnerabilities from these directories, we can just apply the filter to the result after vulnerability detection like https://github.com/aquasecurity/trivy/pull/595.

It doesn't affect the cache. Finally, I realized aquasecurity/trivy#595 was better in terms of cache and let me revert this change.

It's the difference in where we filter.

```
usual:
analysis → cache → detection

skip dirs in fanal:
analysis → filter → cache → detection

skip dirs in trivy:
analysis → cache → detection → filter
```


artifact/local/fs.go
```diff
@@ -13,6 +13,8 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/fanal/analyzer"
+	_ "github.com/aquasecurity/fanal/analyzer/os/alpine"
```
danielpacak, line 16: NIT: We have these imported in cmd/fanal/main.go ; I guess you added the imports for side effects to make test runner happy?

knqyf263: Absolutely. I realized these import should be in fs_test.go and moved them in the previous PR. But I reverted it and they are in fs.go again. I'll fix it in the dedicated PR.

Issue #120, opened 2020-06-10
Feature for scanning OCI Image given tag


shimish2:
I am looking for a feature for scanning an OCI image given an image name and tag. 

I have done the necessary changes, please let me know if I can send an PR for the same and provide further details.

knqyf263:
It would be helpful if you send us the PR!

shimish2:
[https://github.com/aquasecurity/fanal/pull/130](url)

shimish2:
#130

PR #130, opened 2020-08-17
Support for scanning of an OCI image given a tag.

shimish2:
Hi @knqyf263 
This PR will support the feature for scanning an OCI image given a tag, currently it only support first tag.

Condition:- Specify an image name and tag ":" separated.
If correct image name and tag is specified ":" separated, image with given tag will be return otherwise first one will be return.

I have not added test cases for this change, please review this changes and I will add a test code for the same.


shimish2:
#120

image/image.go
```diff
@@ -186,7 +188,21 @@ func fileOpener(fileName string) func() (io.ReadCloser, error) {
 }
 
 func tryOCI(fileName string) (v1.Image, error) {
-	lp, err := layout.FromPath(fileName)
+	var inputTag string
+
+	var inputFileName string
```
knqyf263, line 193: nit
```suggestion
	var inputTag, inputFileName string
```

image/image.go
```diff
@@ -205,11 +221,37 @@ func tryOCI(fileName string) (v1.Image, error) {
...
+					return nil, xerrors.New("invalid OCI image")
+				}
+
+				break
+			}
+		}
+
+		if img == nil {
+			return nil, xerrors.New("invalid OCI image tag")
+		}
```
knqyf263, line 247: Could you define a new function for these processes?

shimish2: Sure, I will define one

shimish2:
@knqyf263 could you please let me know why benchmark is failing and what needs to be done from my side to pass benchmark test, I am not able to run locally due to some tcp timeout issue.

knqyf263:
Hmm. It looks like this change doesn't affect performance. Let me re-run the job.

knqyf263 (approved):

knqyf263:
@shimish2 Thanks!

PR #131, opened 2020-08-19
fix(alpine): add origin package as source package

knqyf263:
In the following example, fanal adds two packages, `sqlite-libs` and `sqlite`. `sqlite` is an origin package of `sqlite-libs`. It is necessary for vulnerability detection because security advisories from Alpine have only information about origin package.

```
C:Q1FGSUbDpfDdWmfKGvkw/BevenRHQ=
P:sqlite-libs
V:3.26.0-r3
A:x86_64
S:481468
I:917504
T:Sqlite3 library
U:http://www.sqlite.org
L:Public-Domain
o:sqlite
m:Carlo Landmeter <clandmeter@gmail.com>
t:1546255353
c:856c64b1dc1f9b8176c60e28808482f8503c4e98
D:so:libc.musl-x86_64.so.1
p:so:libsqlite3.so.0=0.8.6
r:sqlite
F:usr
F:usr/lib
R:libsqlite3.so.0
a:0:0:777
Z:Q1MZwGMEis9uc78EhxmxyozT2ZxZM=
R:libsqlite3.so.0.8.6
a:0:0:755
Z:Q1BOGonxBB2SdxbKENvJNl9ifknaE=
```

But, we should not add the origin package as a stand-alone package since `sqlite` is not installed actually. Also, it will display `sqlite` vulnerabilities, not `sqlite-libs` vulnerabilities. We have to add `sqlite` as `SrcName` and `SrcPackage` of `sqlite-libs` and use them for vulnerability detection like other distributions.


analyzer/pkg/apk/apk.go
```diff
@@ -55,14 +55,9 @@ func (a alpinePkgAnalyzer) parseApkInfo(scanner *bufio.Scanner) (pkgs []types.Pa
...
 		case "o:":
-			origin := string(line[2:])
-			originPkg := types.Package{
-				Name:    origin,
-				Version: version,
-			}
-			if analyzer.CheckPackage(&originPkg) {
-				pkgs = append(pkgs, originPkg)
-			}
+			origin := line[2:]
```
simar7, line 58: Is the `string` cast not necessary?

knqyf263: Yes, `line` was originally `string`. IDE told me that the case is unnecessary 😄

PR #132, opened 2020-08-19
cache: Consolidate Version into one

simar7:
This PR consolidates VersionSchema for Artifact and Blob into one.

This is so we can update this version when the following change:
1. Behavior of analyzer
2. Schema


types/const.go
```diff
@@ -1,6 +1,5 @@
 package types
 
 const (
-	ArtifactJSONSchemaVersion = 1
-	BlobJSONSchemaVersion     = 1
+	SchemaVersion = 1
```
knqyf263, line 4: Do you think we should unify these versions? Also, I think this is no longer `SchemaVersion` because we have to update it when we want to invalidate the cache.

types/image.go
```diff
@@ -60,7 +60,7 @@ type ArtifactReference struct {
 
 // ArtifactInfo is stored in cache
 type ArtifactInfo struct {
-	SchemaVersion int
+	Version       int
```
knqyf263, line 63: I think we should leave a comment about when we have to update the version.

types/const.go
simar7: I think its nice and easy to keep things simple with one version. What's the counter argument?

types/image.go
simar7: What do you propose?

knqyf263:
Close in favor of https://github.com/aquasecurity/fanal/pull/160

PR #134, opened 2020-08-30
refactor(image): minor changes

knqyf263:
ref. https://github.com/aquasecurity/fanal/pull/130

### DIFF
--- a/image/image.go
+++ b/image/image.go
@@ -191,8 +191,9 @@ func tryOCI(fileName string) (v1.Image, error) {
 	var inputTag, inputFileName string
 
 	// Check if tag is specified in input
+	// e.g. /path/to/oci:0.0.1
 	if strings.Contains(fileName, ":") {
-		splitFileName := strings.Split(fileName, ":")
+		splitFileName := strings.SplitN(fileName, ":", 2)
 		inputFileName = splitFileName[0]
 		inputTag = splitFileName[1]
 	} else {
@@ -220,19 +221,7 @@ func tryOCI(fileName string) (v1.Image, error) {
 	}
 
 	// Support image having tag separated by : , otherwise support first image
-
-	if inputTag != "" {
-		return getOCIImage(m, index, inputTag)
-	} else {
-		h := m.Manifests[0].Digest
-
-		img, err := index.Image(h)
-		if err != nil {
-			return nil, xerrors.New("invalid OCI image")
-		}
-
-		return img, nil
-	}
+	return getOCIImage(m, index, inputTag)
 }
 
 func getOCIImage(m *v1.IndexManifest, index v1.ImageIndex, inputTag string) (v1.Image, error) {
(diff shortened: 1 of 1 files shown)


image/image.go
```diff
@@ -220,32 +221,21 @@ func tryOCI(fileName string) (v1.Image, error) {
...
+	return getOCIImage(m, index, inputTag)
 }
 
 func getOCIImage(m *v1.IndexManifest, index v1.ImageIndex, inputTag string) (v1.Image, error) {
 	for _, manifest := range m.Manifests {
 		annotation := manifest.Annotations
 
 		tag := annotation[ispec.AnnotationRefName]
-		if tag == inputTag {
+		if inputTag == "" || // always select the first digest
```
danielpacak, line 232: If I understand this change correctly, such comment to always select the first tag should be place next to the line where we do split imageRef by ":" characters (string.SplitN)

danielpacak: And inputTag is actually firstInputTag isn't it?

knqyf263: OCI can include some tags/digests in a single archive. In that case, `index.json` looks like the following.

```
$ cat /path/to/oci/index.json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:3ec49b236db54454b6b67c52d779c568f2e7de9daab769d327b93561b6307c1a",
      "size": 348,
      "annotations": {
        "org.opencontainers.image.ref.name": "3.11"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:8aaeb7434f78e8c16686a1207078f32f931131e529f10a6cf17585b59c671685",
      "size": 348,
      "annotations": {
        "org.opencontainers.image.ref.name": "3.12"
      }
    }
  ]
}
```

`fanal` used to always select the first digest. In this case, it is `sha256:3ec49b2...` (3.11), but we can specify the tag to be selected now. If a user passes `/path/to/oci:3.12`. we split it and get `3.12`. Then, it looks for the digest with the tag `3.12`. In this case, it is `sha256:8aaeb743...`. If a user doesn't specify a tag such as `/path/to/oci`, it always selects the first digest as before.

In short, it works as follows

- user input: `/path/to/oci` (inputTag is empty)
  - select the first digest (`sha256:3ec49b2...`)
- user input: `/path/to/oci:3.11` (inputTag is `3.11`)
  - select the first digest with the tag `3.11` (`sha256:3ec49b2...`)
- user input: `/path/to/oci:3.12` (inputTag is `3.12`)
  - select the second digest with the tag `3.11` (`sha256:8aaeb743...`)
- user input: `/path/to/oci:3.13` (inputTag is `3.13`) 
  - return an error (no such tag)
- user input: `/path/to/oci:3.13:3.14` (inputTag is `3.13:3.14`) 
  - return an error (no such tag)

A user can specify only one tag.

danielpacak: Got it. Thanks for the detailed explanation!

danielpacak (approved):

PR #135, opened 2020-09-22
Close open connections and files

bastianccm:
Ensure that the client and file is closed either in error cases or once cleanup() is called so no file descriptors are leaked.

If ImageInspectWithRaw
https://github.com/aquasecurity/fanal/blob/1b5f79c38073afe1a11ee19fe1f1095535dc6ce8/image/daemon/image.go#L42
fails the docker client is not closed, and thus will leak filedescriptors to the underlying network connection to docker.
This happens when you reuse the fanal code in a long-running process.

Unfortunately I don't know of a good automated test case, it can be tested by shelling out to lsof to count open files, e.g. in
https://github.com/aquasecurity/fanal/blob/master/image/image_test.go

Example Test
```go
func countOpenFiles() int {
	out, err := exec.Command("/bin/sh", "-c", fmt.Sprintf("lsof -p %v", os.Getpid())).Output()
	if err != nil {
		log.Fatal(err)
	}
	lines := strings.Split(string(out), "\n")
	return len(lines) - 1
}


func TestFdLeak(t *testing.T) {
	for i := 0; i<1000; i++ {
		_, cleanup, err := NewDockerImage(context.Background(), "alpine:3.11", types.DockerOption{})
		cleanup()
		log.Println(err, i, countOpenFiles())
	}
}
```

### DIFF
--- a/image/daemon/image.go
+++ b/image/daemon/image.go
@@ -38,6 +38,11 @@ func Image(ref name.Reference) (v1.Image, func(), error) {
 	if err != nil {
 		return nil, cleanup, xerrors.Errorf("failed to initialize a docker client: %w", err)
 	}
+	defer func() {
+		if err != nil {
+			c.Close()
+		}
+	}()
 
 	inspect, _, err := c.ImageInspectWithRaw(context.Background(), ref.Name())
 	if err != nil {
@@ -50,6 +55,8 @@ func Image(ref name.Reference) (v1.Image, func(), error) {
 	}
 
 	cleanup = func() {
+		c.Close()
+		f.Close()
 		_ = os.Remove(f.Name())
 	}
 


image/daemon/image.go
```diff
@@ -41,15 +41,19 @@ func Image(ref name.Reference) (v1.Image, func(), error) {
 
 	inspect, _, err := c.ImageInspectWithRaw(context.Background(), ref.Name())
 	if err != nil {
+		c.Close()
```
knqyf263, line 44: What if putting `defer` under the error handling of `NewClientWithOpts`? If we add a new error handling, we don't have to add `c.Close()` again.

```
defer func(){
    if err != nil {
        c.Close()
    }
}
```

knqyf263 (commented):
Thank you for the good catch!

image/daemon/image.go
bastianccm: Good idea, much better than repeating the c.Close for every error case. I adopted the change.

knqyf263 (approved):
Thank you so much!

PR #137, opened 2020-10-04
feat: Analyzer for user defined packages.

Johannestegner:
Hi there!
When building docker images, I often build most of the software from source instead of downloading them from a package registry.  
Looking at issues like https://github.com/aquasecurity/trivy/issues/481 in the Trivy repository, make me think that I'm not alone.

This pull request contains a simple analyzer which adds packages to scan from a json file (currently named `/etc/manifest.trivy` open for better naming!). The json file uses the `types.Package` in `types/image.go` as resource in the decoder, although it only requires the name and version to work.

I added a test for it in the `analyzer_test.go` file, and could not find more tests which tests the analyzers, please let me know (if the PR is wanted) if I missed anything when it comes to the tests!

### DIFF
--- a/analyzer/testdata/etc/manifest.trivy
+++ b/analyzer/testdata/etc/manifest.trivy
@@ -1,0 +1,4 @@
+[
+  {"name": "php", "version": "7.4.11" },
+  {"name": "curl", "version": "7.72.0" }
+]
--- a/analyzer/analyzer_test.go
+++ b/analyzer/analyzer_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/aquasecurity/fanal/analyzer"
 	_ "github.com/aquasecurity/fanal/analyzer/library/bundler"
+	_ "github.com/aquasecurity/fanal/analyzer/library/manifest"
 	aos "github.com/aquasecurity/fanal/analyzer/os"
 	_ "github.com/aquasecurity/fanal/analyzer/os/alpine"
 	_ "github.com/aquasecurity/fanal/analyzer/os/ubuntu"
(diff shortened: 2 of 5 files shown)


analyzer/testdata/etc/manifest.trivy
```diff
@@ -0,0 +1,4 @@
+[
```
mrueg, line 1: I would suggest to make this file csv or yaml format, so you can append more easily than with json, wdyt?

Johannestegner: Aye, that is a sane thought, not sure why I went with JSON to be honest, hehe.
Will update that in a bit (after kids are in bed ;) )

Johannestegner: Will have to look a bit deeper into this, it seems like there is no native yaml support, and doesn't feel right to add a new dependency for this tiny feature... I'll figure something out, and I still agree with you that JSON is not the perfect format :)

masahiro331:
I think it's a great idea!
Trivy needs to determine which OS package or which application.
And Trivy needs to select a Driver during vulnerability detection.

e.g. 
https://github.com/aquasecurity/trivy/blob/master/pkg/detector/library/driver.go#L29-L47
https://github.com/aquasecurity/trivy/blob/master/pkg/detector/ospkg/detect.go#L59-L86

In the case of static filenames, it is difficult to sort out the driver.
I find it difficult to sort out the drivers if the file names are static.

Johannestegner:
I think that the most useful thing for it is to look for vulnerabilities in built applications (it will of course require the user to specify the packages with the correct name for the OS distribution, eg: `php8 - 8.0.0_rc2-r0`, but imho, that is quite alright), but making it possible to use the package drivers as well is not a bad idea at all, and I'll look at getting that in too :)

netrounds-fredrik:
This is related to the feature request https://github.com/aquasecurity/trivy/issues/293.

> Trivy needs to determine which OS package or which application.
> And Trivy needs to select a Driver during vulnerability detection.

Regarding what information that the manifest files need to contain, shouldn't it be the same information as passed to the trivy server API (https://github.com/aquasecurity/trivy/blob/master/rpc/detector/service.proto)?

PR #139, opened 2020-10-14
Feat: Created analyzer for NuGet lockfiles.

Johannestegner:
This pull requests include a new analyzer for NuGet `packages.lock.json` files.  
It's basically the same as the other analyzers, but makes use of the NuGet parser which is included in the following `go-dep-parser` [pull request](https://github.com/aquasecurity/go-dep-parser/pull/14).

For reference, feature request issue can be found here: https://github.com/aquasecurity/trivy/issues/681

### DIFF
--- a/analyzer/library/const.go
+++ b/analyzer/library/const.go
@@ -5,6 +5,7 @@ const (
 	Cargo    = "cargo"
 	Composer = "composer"
 	Npm      = "npm"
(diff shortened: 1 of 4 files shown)


knqyf263 (approved):

knqyf263:
Thanks a lot!!

PR #142, opened 2020-11-24
fix(rpm): fill source package information

knqyf263:
The information regarding a source rpm is missing in #141. `rpmcmd` package has the information which is needed for vulnerability detection, but `rpm` package doesn't add the information.
https://github.com/aquasecurity/fanal/blob/702f53856299a4c7dd612682de2e2ea692a4d99a/analyzer/pkg/rpmcmd/rpm.go#L107-L110

We migrated `rpmcmd` to `rpm`, so they need to be the same. In addition, this PR removes `rpmcmd` package which is no longer necessary.


analyzer/pkg/rpm/rpm.go
```diff
@@ -62,21 +67,63 @@ func (a rpmPkgAnalyzer) parsePkgInfo(packageBytes []byte) (pkgs []types.Package,
...
+
+	verIndex := strings.LastIndex(filename[:relIndex], "-")
+	ver = filename[verIndex+1 : relIndex]
+
+	epochIndex := strings.Index(filename, ":")
+	if epochIndex == -1 {
+		epoch = 0
+	} else {
+		epoch, _ = strconv.Atoi(filename[:epochIndex])
+	}
```
simar7, line 121: For my understanding: What's `epochIndex` in the context of a filename? Do you have an example?

analyzer/pkg/rpm/rpm.go
```diff
@@ -62,21 +67,63 @@ func (a rpmPkgAnalyzer) parsePkgInfo(packageBytes []byte) (pkgs []types.Package,
...
 		if arch == "" {
-			arch = "(none)"
+			arch = "None"
 		}
+
+		// parse source rpm
+		var srcName, srcVer, srcRel string
+		if pkg.SourceRpm != "(none)" && pkg.SourceRpm != "" {
+			// source epoch is not included in SOURCERPM
+			srcName, srcVer, srcRel, _, _ = splitFileName(pkg.SourceRpm)
```
simar7, line 77: Since `epoch` is dropped do we still need this logic https://github.com/aquasecurity/fanal/pull/142/files#diff-a25c031570c1f818ffca9c3824028e5731f708802a6ff278b1e6020747bab04cR116-R125

simar7:
We currently have some references in Trivy readme that talk about dependence on `rpm`. Would you remove them as part of a separate Trivy PR?

analyzer/pkg/rpm/rpm.go
knqyf263: You can see an example here.
https://github.com/aquasecurity/fanal/blob/15afdbeacb4bc40e101d9ef12ea1a7ab483ad7ce/analyzer/pkg/rpm/rpm.go#L100

`Epoch` is a number separated by a colon at the beginning of the package name.

knqyf263: But unfortunately `SourceRpm` doesn't include `Epoch`. I don't know why, but I'm guessing it is the same as the epoch of the binary package.

analyzer/pkg/rpm/rpm.go
knqyf263: `splitFilename` is ported from python, so it is a generic function. In this case, `SourceRpm` doesn't include an epoch and we have to use the epoch of the binary package.

knqyf263:
Yes, I'm just working on it in trivy repo.

knqyf263:
I opened the PR in trivy.
https://github.com/aquasecurity/trivy/pull/753

analyzer/pkg/rpm/rpm.go
simar7: I understand your point. But I meant to say that we return epoch here https://github.com/aquasecurity/fanal/blob/15afdbeacb4bc40e101d9ef12ea1a7ab483ad7ce/analyzer/pkg/rpm/rpm.go#L124 

So in our case we don't use the return value could we not remove this logic: https://github.com/aquasecurity/fanal/blob/15afdbeacb4bc40e101d9ef12ea1a7ab483ad7ce/analyzer/pkg/rpm/rpm.go#L117-L121

My apologies if I missed something obvious! 🙇🏻‍♂️

analyzer/pkg/rpm/rpm.go
simar7: I see. thanks!

analyzer/pkg/rpm/rpm.go
knqyf263: You're totally right. We can remove it. I just wanted to have the same functionality as the original function in Python. If they are the same logic between Python and Go, we can easily debug the Go version when it has a bug.
On the other hand, I understand what you mean. It is unnecessary and might be confusing. I've removed it.
https://github.com/aquasecurity/fanal/pull/142/commits/c76ea78da0c290a0ea9d7395421b5df8f551f437

simar7 (approved):
🐻

PR #143, opened 2020-12-06
feat(cache): support Redis

knqyf263:
Close https://github.com/aquasecurity/trivy/issues/765

### Test

```
$ docker run -d --name redis -p 6379:6379 redis:5.0
$ go run cmd/fanal/main.go -cache redis://localhost:6379 image centos:7
centos:7
&{Family:centos Name:7.9.2009}
via image Packages: 148
$ redis-cli
127.0.0.1:6379> keys *
1) "artifact::sha256:8652b9f0cb4c0599575e5a003f5906876e10c1ceb2ab9fe1786712dac14a50cf"
2) "blob::sha256:174f5685490326fc0a1c0f5570b8663732189b327007e47ff13d2ca59673db02"
```


cmd/fanal/main.go
```diff
@@ -62,46 +59,70 @@ func run() (err error) {
...
-				Action:  globalOption(ctx, fsCache, repoAction),
+				Action:  globalOption(ctx, repoAction),
 			},
 		},
 		Flags: []cli.Flag{
 			&cli.BoolFlag{Name: "clear", Aliases: []string{"s"}},
+			&cli.StringFlag{
+				Name:    "cache",
+				Aliases: []string{"c"},
+				Usage:   "cache backend (e.g. redis://localhost:6379)",
```
danielpacak, line 88: Currently it's file system path or Redis URL right? It's out of scope, but I was wondering if we could configure S3 backend in similar way?

cache/redis_test.go
```diff
@@ -0,0 +1,491 @@
...
+				artifactConfig: types.ArtifactInfo{
+					SchemaVersion: 1,
+					Architecture:  "amd64",
+					Created:       time.Date(2020, 11, 14, 0, 20, 4, 0, time.UTC),
+					DockerVersion: "19.03.12",
+					OS:            "linux",
+				},
+			},
+			wantKey:   "fanal::artifact::sha256:8652b9f0cb4c0599575e5a003f5906876e10c1ceb2ab9fe1786712dac14a50cf",
+			wantValue: `{"SchemaVersion":1,"Architecture":"amd64","Created":"2020-11-14T00:20:04Z","DockerVersion":"19.03.12","OS":"linux"}`,
```
danielpacak, line 45: Do we need that? In other places I saw that you're using the value from `args` to assert on expected JSON with `assert.JSONeq()`

cache/redis_test.go
```diff
@@ -0,0 +1,491 @@
...
+		},
+	}
+
+	// Set up Redis test server
+	s, err := miniredis.Run()
+	require.NoError(t, err)
+	defer s.Close()
+
+	// Set key/value pairs
+	v := `{"SchemaVersion":1,"Architecture":"amd64","Created":"2020-11-14T00:20:04Z","DockerVersion":"19.03.12","OS":"linux"}`
```
danielpacak, line 235: If we added AsJson method to ArtifactInfo struc, we could avoid JSON string literals in unit tests:

```go
jsonString, err := ArtifactInfo{}.AsJson()
```

WDYT?

cache/redis_test.go
```diff
@@ -0,0 +1,491 @@
...
+	// Set up Redis test server
+	s, err := miniredis.Run()
+	require.NoError(t, err)
+	defer s.Close()
+
+	t.Run("close", func(t *testing.T) {
+		c := cache.NewRedisCache(&redis.Options{
+			Addr: s.Addr(),
+		})
+		require.NoError(t, c.Close())
```
danielpacak, line 466: This one is actually testing that the `Close()` method did not return any error. Not sure if it's worth it, but it seems that miniredis allows you to check current connections so you could write:

```suggestion
		closeErr := c.Close()
		require.NoError(t, closeErr)
		time.Sleep(3 * time.Second) // give it some time
		assert.Equal(t, 0,s.CurrentConnectionCount(), "The client is disconnected")
```

danielpacak (approved):
Great stuff! I've tested with a bunch of images and it works like a charm. I just left a few minor comments/suggestions.

cache/redis.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return true, missingBlobIDs, nil
+	}
+	if artifactInfo.SchemaVersion != types.ArtifactJSONSchemaVersion {
+		missingArtifact = true
+	}
+	return missingArtifact, missingBlobIDs, nil
+}
+
+func (c RedisCache) Close() error {
```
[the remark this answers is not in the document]
knqyf263: It should be called. I added https://github.com/aquasecurity/fanal/pull/143/commits/f48d6dead75d14b4486fa8e25433f105a6016430

cache/redis.go
```diff
@@ -0,0 +1,140 @@
...
+	if err == redis.Nil {
+		return types.ArtifactInfo{}, xerrors.Errorf("artifact (%s) is missing in Redis cache", artifactID)
+	} else if err != nil {
+		return types.ArtifactInfo{}, xerrors.Errorf("failed to get artifact from the Redis cache: %w", err)
+	}
+
+	var info types.ArtifactInfo
+	err = json.Unmarshal(val, &info)
+	if err != nil {
+		return types.ArtifactInfo{}, xerrors.Errorf("JSON unmarshal error: %w", err)
```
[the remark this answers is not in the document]
knqyf263: Totally. Thanks.

cache/redis.go
```diff
@@ -0,0 +1,140 @@
...
+	val, err := c.client.Get(context.TODO(), key).Bytes()
+	if err == redis.Nil {
+		return types.BlobInfo{}, xerrors.Errorf("blob (%s) is missing in Redis cache", blobID)
+	} else if err != nil {
+		return types.BlobInfo{}, xerrors.Errorf("failed to get blob from the Redis cache: %w", err)
+	}
+
+	var blobInfo types.BlobInfo
+	if err = json.Unmarshal(val, &blobInfo); err != nil {
+		return types.BlobInfo{}, xerrors.Errorf("JSON unmarshal error: %w", err)
```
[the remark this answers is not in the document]
knqyf263: Absolutely. Thanks.

cache/redis.go
```diff
@@ -0,0 +1,140 @@
...
+		missingArtifact = true
+	}
+	return missingArtifact, missingBlobIDs, nil
+}
+
+func (c RedisCache) Close() error {
+	return c.client.Close()
+}
+
+func (c RedisCache) Clear() error {
```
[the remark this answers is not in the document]
knqyf263: My commitment to quality 😆

cmd/fanal/main.go
knqyf263: Yes, we can easily support S3 in the same way. But for now, Aqua SaaS is the only user for S3. If an OSS user requests S3 support via CLI options, we'll do that.

cache/redis_test.go
knqyf263: `json.Marshal` is used in `PutArtifact` / `PutBlob` and I'm afraid that the test passes even though we don't know what `json.Marshal` returns. If `json.Marshal` doesn't work as expected, we may not be able to test `PutArtifact` properly 

Let's say we want to test a function for rounding up.

```
func roundUp(a float64) float64 {
    return math.Floor(a)
```

But unfortunately, we mistakenly used `math.Floor`, not `math.Ceil`. The unit test might be as follows.

```
func TestRoundUp(t *testing.T) {
	input := 3.2
	got := roundUp(input)
	want := math.Floor(input) // should be 4.0, but 3.0 actually
	assert.Equal(t, want, got) // pass => want: 3.0, got: 3.0
}
```
 
The above test will pass even if there is a bug since `math.Floor` is also used in the test by mistake. The test should be

```
func TestRoundUp(t *testing.T) {
	input := 3.2
	got := roundUp(input)
	want := 4.0
	assert.Equal(t, want, got) // fail => want: 4.0, got: 3.0
}
```

The first test looks useless and I'm feeling like it's similar if we use `json.Marshal` in the unit test.

```
func (c RedisCache) PutBlob(blobID string, blobInfo types.BlobInfo) error {
        ...
	b, err := json.Marshal(blobInfo)
        ...
	if err := c.client.Set(context.TODO(), key, string(b), 0).Err(); err != nil {
        ...
}

func TestRedisCache_PutBlob(t *testing.T) {
	input := types.ArtifactInfo{SchemaVersion: 1}
	got :=c.PutArtifact(input)
	want := json.Marshal(input)
	assert.Equal(t, want, got)
```

I omitted some arguments/return values, but hopefully, it makes sense to you.
The test might pass even if `json.Marshal` always returns an empty value or something wrong as we don't expect.

So, I think the wanted value basically should be literal.

```
func TestRedisCache_PutBlob(t *testing.T) {
	input := types.ArtifactInfo{SchemaVersion: 1}
	got :=c.PutArtifact(input)
	want := `{"SchemaVersion":1}`
	assert.Equal(t, want, got)
```

We know the behavior of `json.Marshal` well, so the test probably works properly, but I feel it is basically doing the same thing as the example of `roundUp`.

But as you said, the string literal makes it difficult to maintain. I also don't want to use the string literal representing JSON. Actually, `PutBlob` had too long JSON and that is why I used `json.Marshal` in `TestRedisCache_PutBlob`. I don't think it is the best way.
Any better ideas?

cache/redis_test.go
knqyf263: I think we need string literals to test `AsJson` anyway.

danielpacak: Thanks for the detailed explanation. I was mainly wondering why in some cases there are string literals and structs in other places. Other than that I don't have strong opinions about string literals in unit tests as long as they do the job.

The way I read the code is that you're trying to test 2 things: 1) The usage of Go and Redis client so values are stored as JSON 2) JSON marshalling/unmarshaling

If you testes marshalling/unmarshalling somewhere else (which might be overhead) you could probably unit test Redis code with structs only. But then you'll need another integration test that tests both Redis and JSON stuff.

knqyf263: Thanks for summarizing. It totally makes sense. I wanted to test 1) and 2) both. But eventually, I switched to structs. https://github.com/aquasecurity/fanal/pull/143/commits/0d842dfea762f3d8241973e4f93f3869f82a7cdd

Mixing string literals and structs in unit tests might be confusing as you said. In addition, it looks enough to test 1) only as we already know JSON marshaling/unmarshaling well.

knqyf263:
@danielpacak I updated the test a little. Could you take a look last time?

danielpacak (approved):
LGMT! Great discussion about testing Redis client and JSON 🚀

PR #149, opened 2021-01-06
feat(image): support Podman

knqyf263:
## Abstract
Podman provides Go bindings, but it makes cross-compile difficult since it depends on CGO. It can be compiled for Linux, but difficult for other platforms.
https://podman.io/blogs/2020/08/10/podman-go-bindings.html

This time, I simply developed our own Podman client to use the REST API via HTTP against the Podman socket such as `/run/user/${UID}/podman`.

To use it with Podman, a user needs to activate the podman socket.

```
$ systemctl --user start podman.socket
```

For more details, see here.
https://podman.io/blogs/2020/08/10/podman-go-bindings.html#start-the-podman-system-service-

## API reference
https://docs.podman.io/en/latest/_static/api.html

## Issue
https://github.com/aquasecurity/trivy/issues/792


masahiro331:
:shipit:

masahiro331 (approved):

PR #151, opened 2021-01-11
feat(analyzer): support Red Hat build info

knqyf263:
Red Hat official images have content manifests under `/root/buildinfo`. It includes content sets that can be converted into CPE and it is necessary to query appropriate OVAL feeds.
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index

If content manifests don't exist in a container image, the build information needs to be extracted from `/root/buildinfo/Dockerfile-*`. Those values are converted to CPE through Pyxis API.
https://catalog.redhat.com/api/containers/v1/ui/#/

Eventually, each package needs to have content sets.


analyzer/analyzer.go
```diff
@@ -85,6 +96,27 @@ func (r *AnalysisResult) Merge(new *AnalysisResult) {
 	if len(new.Applications) > 0 {
 		r.Applications = append(r.Applications, new.Applications...)
 	}
+
+	if new.BuildInfo != nil {
+		r.BuildInfo = new.BuildInfo
```
masahiro331, line 101: Shouldn't this buildInfo be merged in?

tonaim (approved):

PR #153, opened 2021-02-10
feat(analyzer): disable analyzers

knqyf263:
This PR makes it possible to disable analyzers.


masahiro331:
IMO.
When sharing cache in server mode, The disable-analyzer can have an unexpected negative cache .
For example, John  use trivy command with removed-packages option, after that Fred use trivy command.
At this point Fred receives the result of running removed-pkgs.

artifact/remote/git.go
```diff
@@ -38,7 +40,10 @@ func NewArtifact(rawurl string, c cache.ArtifactCache) (artifact.Artifact, func(
 		_ = os.RemoveAll(tmpDir)
 	}
 
-	return local.NewArtifact(tmpDir, c), cleanup, nil
+	// JAR/WAR/EAR doesn't need to be analyzed in git repositories.
+	disabled = append(disabled, analyzer.TypeJar)
```
masahiro331, line 44: Just my preference:
I think it is better to implement this process in Trivy.

knqyf263: Some analyzers should not be enabled regardless of callers. I think it is better to hard-code and disable them here.

knqyf263:
@masahiro331 Nice catch. The change would be large, so I'll work on it in a dedicated PR.

artifact/remote/git.go
masahiro331: I got it.

masahiro331:
LGTM! :shipit:

masahiro331 (approved):

PR #156, opened 2021-02-23
feat(cache): add analyzer versions

knqyf263:
## Overview
Add analyzer versions to cache like the following:

```
"AnalyzerVersions": {
    "alpine": 1,
    "debian": 2,
    "redhat": 1,
    "python": 1,
    "ruby": 3,
}
```

Note that all enabled analyzers are stored. Even if you scan Alpine Linux, "debian", "redhat" and other OS analyzers will be stored if they are enabled.

If one of those versions is older than the analyzer versions with newly created artifact/blog information, this cache will be invalidated. Also, if one of the enabled analyzers is missing in the cache, the cache will be invalidated.

Let's say the following versions are cached.

```
"AnalyzerVersions": {
    "alpine": 1,
    "ruby": 3,
}
```

If the new values are like the following, the above cache will be replaced with the new one.

```
"AnalyzerVersions": {
    "alpine": 2,
    "ruby": 3,
}
```

Also, if a new analyzer missing in the cache is added or enabled, the cache will be invalidated.

```
"AnalyzerVersions": {
    "alpine": 2,
    "ruby": 3,
    "java": 1,
}
```

In this case, the "java" analyzer is missing in the cache.

On the other hand, the following case doesn't invalidate the cache since the "alpine" analyzer was enabled in the cache, even though the "ruby" analyzer is not enabled this time. The larger, the better

```
"AnalyzerVersions": {
    "alpine": 2,
}
```


cache/cache.go
```diff
@@ -44,3 +45,19 @@ type LocalArtifactCache interface {
...
 }
+
+// isStale checks if the cache is stale or not.
+// e.g. When {"alpine": 1, "python": 2} is cached and {"alpine": 2, "python": 2} is sent, the cache is stale.
+// Also, {"python": 2} cache must be replaced by {"alpine": 1, "python": 2}.
+func isStale(current, cached map[string]int) bool {
+	for analyzerType, currentVersion := range current {
+		cachedVersion, ok := cached[analyzerType]
+		if !ok {
+			return true
```
masahiro331, line 56: I think this line is return false?

cache/cache.go
```diff
@@ -44,3 +45,19 @@ type LocalArtifactCache interface {
 	// Clear deletes the local database
 	Clear() (err error)
 }
+
+// isStale checks if the cache is stale or not.
+// e.g. When {"alpine": 1, "python": 2} is cached and {"alpine": 2, "python": 2} is sent, the cache is stale.
+// Also, {"python": 2} cache must be replaced by {"alpine": 1, "python": 2}.
+func isStale(current, cached map[string]int) bool {
+	for analyzerType, currentVersion := range current {
+		cachedVersion, ok := cached[analyzerType]
```
masahiro331, line 54: In this case, can't determine stale.
The isStale `return false` .
```
current: map[ amazon:1 ]
cached: map[ amazon:1, alpine:1 ]
```

knqyf263: It is intended as it is explained in the description.

masahiro331: Thanks, make sense.

masahiro331 (approved):

analyzer/command/apk/apk.go
```diff
@@ -273,3 +275,7 @@ func (a alpineCmdAnalyzer) Required(targetOS types.OS) bool {
 func (a alpineCmdAnalyzer) Type() analyzer.Type {
 	return analyzer.TypeApkCommand
 }
+
+func (a alpineCmdAnalyzer) Version() int {
+	return 1
```
masahiro331, line 280: It is better to write in const.

```
analyzer/os/redhatbase/oracle.go
analyzer/os/redhatbase/fedora.go
analyzer/os/redhatbase/centos.go
analyzer/os/redhatbase/redhatbase.go
analyzer/command/apk/apk.go
```

PR #158, opened 2021-02-24
feat(config): support HCL files

ronaudinho:
- [x] HCL1
- [x] HCL2 based on [conftest](https://github.com/open-policy-agent/conftest)
- [ ] HCL2 based on [hcl v2](https://github.com/hashicorp/hcl)


analyzer/config/hcl2/hcl2_test.go
```diff
@@ -0,0 +1,108 @@
...
+						Type:     config.HCL2,
+						FilePath: "testdata/deployment.hcl",
+						Content: map[string]interface{}{
+							"apiVersion": "apps/v1",
+							"kind":       "Deployment",
+							"metadata": map[string]interface{}{
+								"name": "hello-kubernetes",
+							},
+							"spec": map[string]interface{}{
+								"replicas": float64(3),
```
ronaudinho, line 38: [conftest hcl2 parser](https://github.com/open-policy-agent/conftest/tree/master/parser/hcl2) does not use [hashicorp's library](https://github.com/hashicorp/hcl), which results in this inaccurate unmarshalling. for our purpose, i think it does not really matter, but do you think we should try to use the native library instead?

analyzer/config/hcl2/hcl2.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/aquasecurity/fanal/types"
+)
+
+func init() {
+	analyzer.RegisterAnalyzer(&hcl2ConfigAnalyzer{
+		parser: &hcl2.Parser{},
+	})
+}
+
+const version = 2
```
ronaudinho, line 21: does version to analyzer's version? I might need to change this.

ronaudinho:
I combined HCL1 and HCL2 config analyzer into one, let me know if you prefer [previous work](https://github.com/aquasecurity/fanal/pull/158/commits/5ecb592e638fb6137344b4eb526b25d3508e5875)

analyzer/config/hcl2/hcl2.go
knqyf263: Yes, it should be an analyzer's version. Start from 1.

analyzer/config/hcl2/hcl2_test.go
knqyf263: What kind of differences are there? `float64(3)` looks fine.

knqyf263: `hcl2json` seems to use `hashicorp/hcl/v2` internally.
https://github.com/tmccombs/hcl2json/blob/main/convert/convert.go#L8

ronaudinho: it's just that given this HCL config
```
spec {
  replicas = 3
}
```
the current parser will give
```
"spec": map[string]interface{}{
  "replicas": float64(3),
},
```
instead of
```
"spec": map[string]interface{}{
  "replicas": int(3), // or any other int type
},
```
which is probably not a big deal for what we're trying to do

knqyf263: I'm wondering it is difference between hashicorp/hcl/v1 and hashicorp/hcl/v2, not between tmccombs/hcl2json/convert and hashicorp/hcl/v2.

knqyf263: If `tmccombs/hcl2json/convert` results in inaccurate results, we should replace it with `hashicorp/hcl/v2`, but the library internally uses `hashicorp/hcl/v2`. So, I feel like `float64(3)` is a correct result. Could you parse the test file by `hashicorp/hcl/v2` and check if it returns int(3) or float64(3)?

ronaudinho: FWIW, the spec mentions [this](https://github.com/hashicorp/hcl/blob/main/json/spec.md#numbers) and [this](https://github.com/hashicorp/hcl/blob/main/spec.md#primitive-types)

knqyf263: Hmm. I wanted to look into it, but it doesn't matter in our case actually. If we find an issue in the future, we can fix it.

knqyf263 (approved):

PR #159, opened 2021-02-24
feat(config): support JSON files

ronaudinho:

### DIFF
--- a/analyzer/config/json/json.go
+++ b/analyzer/config/json/json.go
@@ -1,0 +1,55 @@
+package json
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/open-policy-agent/conftest/parser/json"
+	"golang.org/x/xerrors"
+
+	"github.com/aquasecurity/fanal/analyzer"
+	"github.com/aquasecurity/fanal/analyzer/config"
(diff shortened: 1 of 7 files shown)


analyzer/config/json/json.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, required := range requiredExts {
+		if ext == required {
+			return true
+		}
+	}
+	return false
+}
+
+func (a jsonConfigAnalyzer) Type() analyzer.Type {
+	return analyzer.TypeYaml
```
knqyf263, line 54:
```suggestion
	return analyzer.TypeJSON
```

knqyf263: I didn't add any test of `Type()` in the YAML analyzer, but we might need to add a pretty simple test.

knqyf263 (changes requested):

analyzer/config/json/json.go
ronaudinho: added test for `Type()` here. I just realized that `Version()` does not have test too, would it be necessary?

knqyf263 (approved):
:shipit:

PR #160, opened 2021-02-24
feat(cache): introduce versioned keys

knqyf263:
This PR adds analyzer versions to cache keys as suffixes as follows.

```
Before: sha256:24df0d4e20c0f42d3703bf1f1db2bdd77346c7956f74f423603d651e8e5ae8a7
After:  sha256:24df0d4e20c0f42d3703bf1f1db2bdd77346c7956f74f423603d651e8e5ae8a7/1210112203101
```

`1210112203101` represents [Alpine version][Amazon version][Debian version]...

If even one analyzer version is different, it will be saved as a different cache key.


analyzer/analyzer.go
```diff
@@ -135,21 +137,34 @@ func NewAnalyzer(disabledAnalyzers []Type) Analyzer {
...
+// e.g. alpine: 1, amazon: 3, debian: 2 => 132
+// When the amazon analyzer is disabled => 102
+func (a Analyzer) AnalyzerVersions() string {
+	var versions string
+	for _, driver := range analyzers {
+		if isDisabled(driver.Type(), a.disabled) {
+			versions += "0"
+			continue
+		}
+		versions += fmt.Sprint(driver.Version())
```
masahiro331, line 154: The implementation of AnalyzerVersions depends on the import order.
This is uniquely determined by `go fmt` , but may be affected by changes in package names, etc.

There are several ways to do solution, I can't offer a good this... 
I think we should leave a comment here and Trivy for now.

masahiro331 (approved):

masahiro331:
LGTM!!

PR #161, opened 2021-02-25
feat(config): support Dockerfiles

ronaudinho:


analyzer/config/docker/docker.go
```diff
@@ -0,0 +1,65 @@
...
+var requiredExts = []string{"Dockerfile"}
+
+type dockerConfigAnalyzer struct {
+	parser *docker.Parser
+}
+
+func (a dockerConfigAnalyzer) Analyze(target analyzer.AnalysisTarget) (*analyzer.AnalysisResult, error) {
+	var parsed interface{}
+	if err := a.parser.Unmarshal(target.Content, &parsed); err != nil {
+		return nil, xerrors.Errorf("unable to parse Docker (%s): %w", target.FilePath, err)
```
knqyf263, line 33:
```suggestion
		return nil, xerrors.Errorf("unable to parse Dockerfile (%s): %w", target.FilePath, err)
```

analyzer/config/docker/docker.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Required returns true if base filepath contains of the requiredExts
+// case-insensitive
+func (a dockerConfigAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	base := filepath.Base(filePath)
+	for _, required := range requiredExts {
+		if strings.EqualFold(base, required) {
+			return true
+		}
+		if strings.Contains(strings.ToLower(base), strings.ToLower(required)) {
```
knqyf263, line 52: Let's start with `HasPrefix`. There might be "generate-Dockerfile.sh" or something like that.

```suggestion
		if strings.HasPrefix(strings.ToLower(base), strings.ToLower(required)) {
```

analyzer/config/docker/docker.go
```diff
@@ -0,0 +1,65 @@
...
+
+func init() {
+	analyzer.RegisterAnalyzer(&dockerConfigAnalyzer{
+		parser: &docker.Parser{},
+	})
+}
+
+const version = 1
+
+var requiredExts = []string{"Dockerfile"}
```
knqyf263, line 24: We don't have a plan to add more extensions for now. Also, it is not an extension.

```suggestion
var requiredFile = "Dockerfile"
```

analyzer/config/docker/docker.go
```diff
@@ -0,0 +1,65 @@
...
+			Content:  parsed,
+		}},
+	}, nil
+}
+
+// Required returns true if base filepath contains of the requiredExts
+// case-insensitive
+func (a dockerConfigAnalyzer) Required(filePath string, _ os.FileInfo) bool {
+	base := filepath.Base(filePath)
+	for _, required := range requiredExts {
```
knqyf263, line 48: Fix here according to the previous comment.

analyzer/config/docker/docker_test.go
```diff
@@ -0,0 +1,184 @@
...
+									"Value":  []interface{}{"echo hello"},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:      "happy path with multi-step",
```
knqyf263, line 61: If I understand correctly, it is "multi-stage` rather than "multi-step", right?

```suggestion
			name:      "happy path with multi-stage",
```

analyzer/config/docker/docker_test.go
```diff
@@ -0,0 +1,184 @@
...
+			inputFile: "testdata/Dockerfile.multistep",
```
knqyf263, line 62: ditto

knqyf263 (changes requested):
Thanks. Would you fix the conflict?

analyzer/config/docker/docker.go
ronaudinho: that makes sense. I added having `Dockerfile` as extension condition as well, seems like it is quite common i.e. https://github.com/docker/docker.github.io/issues/9245

analyzer/config/docker/docker.go
ronaudinho: fixed.

analyzer/config/docker/docker.go
ronaudinho: fixed.

analyzer/config/docker/docker_test.go
ronaudinho: fixed.

analyzer/config/docker/docker_test.go
ronaudinho: fixed.

analyzer/config/docker/docker.go
```diff
@@ -0,0 +1,69 @@
...
+	base := filepath.Base(filePath)
+	if strings.EqualFold(base, requiredFile) {
+		return true
+	}
+	if strings.HasPrefix(strings.ToLower(base), strings.ToLower(requiredFile)) {
+		return true
+	}
+
+	ext := filepath.Ext(filePath)
+	if strings.EqualFold(strings.TrimLeft(ext, "."), requiredFile) {
```
knqyf263, line 56: `abcdDockerfile` also matches this condition. I prefer to compare with `.dockerfile`.

```suggestion
	if strings.EqualFold(ext, "." + requiredFile) {
```

knqyf263 (commented):
Could you address the conflict?

analyzer/config/docker/docker.go
```diff
@@ -0,0 +1,65 @@
...
+			Type:     config.Dockerfile,
+			FilePath: target.FilePath,
+			Content:  parsed,
+		}},
+	}, nil
+}
+
+// Required does a case-insensitive check for filePath and returns true if
+// filePath equals/startsWith/hasExtension requiredFile
+func (a dockerConfigAnalyzer) Required(filePath string, _ os.FileInfo) bool {
```
ronaudinho, line 46: I changed the entire checks to accept only `dockerfile.*` or `*.dockerfile`, with added test cases

knqyf263 (approved):
LGTM 👍

PR #163, opened 2021-03-05
fix execution on Windows

### LINKED ISSUE #125: Support for execution on Windows
Since most of the implementation of this is platform independent (pulling images from repositories and extracting them), it should be possible to execute this on Windows.  Currently the main problem is that paths being fed into the analyzers are in Windows form (i.e. backslash), while the analyzers are looking for key files using UNIX path separators.

Need to either have the analyzers look for paths using the native separator, or have the paths fed to the analyzers consistently use UNIX path separators.

ronaudinho:
resolves #125

fix `filepath.Clean` and `filepath.Join` on Windows

manually tested for:
- [x] `fanal img`
- [x] `fanal ar`
- [x] `fanal fs`
- [x] `fanal repo`

### TODO
- [ ] fix disabled unit tests
  - [ ] [artifact/image/image_windows_test.go](https://github.com/ronaudinho/fanal/blob/b1d2216c212f20c675a3eaf7003169b648403b67/artifact/image/image_windows_test.go)
  - [ ] [artifact/local/fs_windows_test.go](https://github.com/ronaudinho/fanal/blob/b1d2216c212f20c675a3eaf7003169b648403b67/artifact/local/fs_windows_test.go)
  - [ ] [artifact/remote/git_test.go](https://github.com/ronaudinho/fanal/blob/b1d2216c212f20c675a3eaf7003169b648403b67/artifact/remote/git_windows_test.go)
  - [ ] [image/daemon/podman_test.go](https://github.com/ronaudinho/fanal/blob/b1d2216c212f20c675a3eaf7003169b648403b67/image/daemon/podman_test.go)
- [ ] fix integration tests for CI

#### Miscellaneous
- `fanal fs` and `fanal repo` yields different `ArtifactDetail.Applications` even though they share the same `Inspect` function

### DIFF
--- a/image/daemon/podman_test.go
+++ b/image/daemon/podman_test.go
@@ -1,3 +1,5 @@
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
+
 package daemon
 
 import (
--- a/walker/walk.go
+++ b/walker/walk.go
@@ -2,6 +2,7 @@ package walker
 
 import (
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/aquasecurity/fanal/analyzer"
@@ -19,6 +20,7 @@ type WalkFunc func(filePath string, info os.FileInfo, opener analyzer.Opener) er
 
 func isIgnored(filePath string) bool {
 	filePath = strings.TrimLeft(filePath, "/")
+	filePath = filepath.Clean(filePath)
 	for _, path := range strings.Split(filePath, utils.PathSeparator) {
 		if utils.StringInSlice(path, library.IgnoreDirs) {
 			return true
--- a/walker/walk_windows_test.go
+++ b/walker/walk_windows_test.go
@@ -1,5 +1,3 @@
-// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
-
 package walker
 
 import (
@@ -32,8 +30,13 @@ func Test_isIgnore(t *testing.T) {
 			want: true,
 		},
 		{
-			name: "ignore dir in dirs",
+			name: "ignore dir in dirs slash separator",
 			dirs: []string{"foo/.git", "foo/node_modules"},
+			want: true,
+		},
+		{
+			name: "ignore dir in backslash separator",
+			dirs: []string{`foo\.git`, `foo\node_modules`},
 			want: true,
 		},
 		{
--- a/.github/workflows/windows_test.yml
+++ b/.github/workflows/windows_test.yml
@@ -1,0 +1,16 @@
+name: Windows Test
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  unittest:
+    name: Unit Test
+    runs-on: windows-latest
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1 
+
+    - name: Run unit tests
+      run: go test ./...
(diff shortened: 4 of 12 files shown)


image/daemon/podman_test.go
```diff
@@ -1,3 +1,5 @@
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
```
ronaudinho, line 1: disabled for CI purpose

ronaudinho:
should we move **TODO** to its own issues? @knqyf263

walker/walk.go
```diff
@@ -19,6 +20,7 @@ type WalkFunc func(filePath string, info os.FileInfo, opener analyzer.Opener) er
 
 func isIgnored(filePath string) bool {
 	filePath = strings.TrimLeft(filePath, "/")
+	filePath = filepath.Clean(filePath)
```
knqyf263, line 23: What about inserting `filepath.ToSlash` before L22? In that case, we probably need to replace `utils.PathSeparator` with `/`.

```
	filePath = filepath.ToSlash(filePath)
        filePath = strings.TrimLeft(filePath, "/")
	filePath = filepath.Clean(filePath)
```

walker/walk_test.go
```diff
@@ -9,19 +11,49 @@ import (
...
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.want {
+				for _, fp := range tt.dirs {
+					assert.True(t, isIgnored(fp))
+				}
+			} else {
+				for _, fp := range tt.dirs {
+					assert.False(t, isIgnored(fp))
+				}
+			}
```
knqyf263, line 56:
```suggestion
			for _, fp := range tt.dirs {
				assert.Equal(t, tt.want, isIgnored(fp))
			}
```

walker/walk_test.go
```diff
@@ -1,3 +1,5 @@
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
```
knqyf263, line 1: The following tests look the same as Windows. So, we don't need this build tag, right? In other words, do these tests pass even on Windows?

walker/walk_windows_test.go
```diff
@@ -0,0 +1,62 @@
...
+		{
+			name: "ignore dir in backslash separator",
+			dirs: []string{`foo\.git`, `foo\node_modules`},
+			want: true,
+		},
+		{
+			name: "not ignore dirs",
+			dirs: []string{"foo", "foo/bar"},
+			want: false,
+		},
```
knqyf263, line 46: We should keep only the difference.

```suggestion
		{
			name: "ignore dir in backslash separator",
			dirs: []string{`foo\.git`, `foo\node_modules`},
			want: true,
		},

```

image/daemon/podman_test.go
knqyf263, line 1:
```suggestion
// +build !windows
```

knqyf263:
> testing.go:915: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestFSCache_PutArtifact_happy_path562076382\001\fanal\fanal.db: The process cannot access the file because it is being used by another process.

The test is failing currently due to the above error. We may have forgotten to close the DB.

ronaudinho:
> > testing.go:915: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestFSCache_PutArtifact_happy_path562076382\001\fanal\fanal.db: The process cannot access the file because it is being used by another process.
> 
> The test is failing currently due to the above error. We may have forgotten to close the DB.

This did not happen when I ran the test on my local. I'll try to repro and find out the cause.

PR #164, opened 2021-03-19
test: support config

ronaudinho:
### added tests
- analyzer/config/scanner/scanner_test

### updated tests
- adjust returns in analyzer/config/...
- updates `NewArtifact` calls in artifact/...

### misc
- wrap errors

### need feedbacks
- many duplicates testdata in analyzer/config/.../testdata, do we need to test rego for each format? or perhaps there is a better way to do it
- hashes change intended? most of the updates on artifact/... tests are incorporating new hashes for blobs


artifact/image/image_test.go
```diff
@@ -25,10 +26,11 @@ import (
 	depTypes "github.com/aquasecurity/go-dep-parser/pkg/types"
 )
 
-func TestArtifact_Inspect(t *testing.T) {
+func Test_imageArtifact_Inspect(t *testing.T) {
```
ronaudinho, line 29: changed to make it easier to trace hashes change, I forgot to change it back :facepalm:

artifact/local/fs_test.go
```diff
@@ -9,17 +9,19 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/aquasecurity/fanal/analyzer"
+	"github.com/aquasecurity/fanal/analyzer/config"
 	"github.com/aquasecurity/fanal/cache"
 	"github.com/aquasecurity/fanal/types"
 )
 
-func TestArtifact_Inspect(t *testing.T) {
+func Test_localArtifact_Inspect(t *testing.T) {
```
ronaudinho, line 17: same as above, changed to make it easier to trace hashes change, I forgot to change it back :facepalm:

artifact/image/image_test.go
```diff
@@ -104,70 +106,61 @@ func TestArtifact_Inspect(t *testing.T) {
...
+								types.PackageInfo{
 									FilePath: "var/lib/dpkg/status.d/base",
 									Packages: []types.Package{
-										{
-											Name:       "base-files",
-											Version:    "9.9+deb9u9",
-											SrcName:    "base-files",
-											SrcVersion: "9.9+deb9u9",
-										},
+										{Name: "base-files", Version: "9.9+deb9u9", SrcName: "base-files", SrcVersion: "9.9+deb9u9"},
```
ronaudinho, line 151: cosmetic change, to match the rest of the test

artifact/image/image_test.go
```diff
@@ -352,20 +370,20 @@ func TestArtifact_Inspect(t *testing.T) {
 							PackageInfos: []types.PackageInfo{{
 								FilePath: "lib/apk/db/installed",
 								Packages: []types.Package{
-									{Name: "musl", Version: "1.1.24-r2", SrcName: "musl", SrcVersion: "1.1.24-r2"},
-									{Name: "busybox", Version: "1.31.1-r9", SrcName: "busybox", SrcVersion: "1.31.1-r9"},
 									{Name: "alpine-baselayout", Version: "3.2.0-r3", SrcName: "alpine-baselayout", SrcVersion: "3.2.0-r3"},
 									{Name: "alpine-keys", Version: "2.1-r2", SrcName: "alpine-keys", SrcVersion: "2.1-r2"},
+									{Name: "apk-tools", Version: "2.10.4-r3", SrcName: "apk-tools", SrcVersion: "2.10.4-r3"},
```
ronaudinho, line 375: should be intended effect of this [sort](https://github.com/aquasecurity/fanal/blob/d6b8630898e62792c2638b5b64f4e4e05a258327/artifact/image/image.go#L179)

analyzer/config/docker/testdata/deny.rego
```diff
@@ -0,0 +1,13 @@
...
+  "foo"
+]
+
+deny[msg] {
+  input[i].Cmd == "from"
+  val := input[i].Value
+  contains(val[i], denylist[_])
+
+  msg = sprintf("deny: image found %s", [val])
+}
```
knqyf263, line 13: We need to return structured messages as follows. But even if a rule returns only "msg", we have to handle it. So, both cases are necessary. We can keep `violation.rego` as is and change this rego to return a structured message. Also, we might have the case missing some fields like `{"msg": "foo", "id": "RULE-100"}`. In this case, `type` and `severity` are missing.

```suggestion
deny[res] {
  input[i].Cmd == "from"
  val := input[i].Value
  contains(val[i], denylist[_])

  res = {"type": "Docker Security Check", "msg": sprintf("deny: image found %s", [val]), "severity": "HIGH", "id": "RULE-100"}
}
```

analyzer/config/hcl/testdata/deny.rego
```diff
@@ -0,0 +1,7 @@
+package testdata
+
+deny[msg] {
+    rpl = input.spec[_].replicas
+	rpl > 2
+    msg = sprintf("deny: too many replicas: %d", [rpl])
```
knqyf263, line 6: ditto

analyzer/config/hcl/testdata/deny.rego
```diff
@@ -0,0 +1,7 @@
+package testdata
+
+deny[msg] {
+    rpl = input.spec[_].replicas
+	rpl > 2
```
knqyf263, line 5: Looks like only this line has different indents.

```suggestion
    rpl > 2
```

analyzer/config/hcl/testdata/non.rego
```diff
@@ -0,0 +1,7 @@
+package testdata
+
+deny[msg] {
+    rpl = input.spec[_].replicas
+	rpl > 3
```
knqyf263, line 5:
```suggestion
    rpl > 3
```

analyzer/config/hcl/testdata/violation.rego
```diff
@@ -0,0 +1,7 @@
+package testdata
+
+violation[{"msg": msg, "details":{}}] {
+    rpl = input.spec[_].replicas
+	rpl > 2
```
knqyf263, line 5:
```suggestion
    rpl > 2
```

analyzer/config/hcl/testdata/warn.rego
```diff
@@ -0,0 +1,7 @@
+package testdata
+
+warn[msg] {
+    rpl = input.spec[_].replicas
+	rpl > 2
```
knqyf263, line 5:
```suggestion
    rpl > 2
```

analyzer/library/const.go
```diff
@@ -0,0 +1,5 @@
+package library
+
+const (
+	Jar = "jar"
```
knqyf263, line 4: I moved the value under `types`.
https://github.com/aquasecurity/fanal/blob/support_config/types/const.go#L18

artifact/image/image_test.go
knqyf263: Will you change it back again? Some Go tools align the convention like TestStructName_MethodName. It allows us to jump to test code easily.

artifact/image/image_test.go
knqyf263: Yes, thanks.

cache/key_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/aquasecurity/fanal/analyzer/config"
 )
 
 func TestWithVersionSuffix(t *testing.T) {
```
knqyf263, line 12:
```suggestion
func TestCalcKey(t *testing.T) {
```

cache/key_test.go
```diff
@@ -12,47 +15,28 @@ func TestWithVersionSuffix(t *testing.T) {
...
-				versioned: "sha256:5c534be56eca62e756ef2ef51523feda0f19cd7c15bb0c015e3d6e3ae090bf6e/111101112110013",
-			},
-			want: "sha256:5c534be56eca62e756ef2ef51523feda0f19cd7c15bb0c015e3d6e3ae090bf6e",
+			want: "sha256:1acfb4a0eb0acb0b1e8d462df18841581b960d5df190b32f3327df4820e5bc7b",
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := TrimVersionSuffix(tt.args.versioned)
+			got, err := CalcKey(tt.args.key, tt.args.version, &config.ScannerOption{})
```
knqyf263, line 34: Could you add cases where ScannerOption is not empty? I want to make sure that the key will be different if we pass `FilePatterns` and `PolicyPaths`. Also, I'd like to check if the key is the same when we pass `[]string{}`, not nil.

knqyf263 (commented):
>many duplicates testdata in analyzer/config/.../testdata, do we need to test rego for each format? or perhaps there is a better way to do it

If there are some duplicate testdata, you can put them under `analyzer/config`. It would be `analyzer/config/testdata` and each test case can load `../testdata/non.rego`.

artifact/image/image_test.go
ronaudinho: yup will do

analyzer/config/docker/testdata/deny.rego
ronaudinho: alright will try to add more cases like this.

ronaudinho:
> If there are some duplicate testdata, you can put them under `analyzer/config`. It would be `analyzer/config/testdata` and each test case can load `../testdata/non.rego`.

I think this would be a better approach, will update the PR with this

analyzer/config/hcl/testdata/deny.rego
ronaudinho: not really sure why, the indent looks the same in my editor :thinking: anyway I'll fix this as well

analyzer/library/const.go
ronaudinho: ah no wonder, then I'll delete this and use the one from `types`

cache/key_test.go
ronaudinho: okay, I was thinking of adding cases to this but wanted to make sure if everything else was on the right track first.

knqyf263: Looks the right way 👍 . Let me describe the background. IaC config files might include sensitive information such as passwords and API keys. If we store them in the cache, it might increase the attack surface for attackers. So, I changed it not to store file contents in the cache. Instead, only the results are stored in the cache now. But let's say we scan config files in container images. If a user updates Rego policies, the result should be updated. We must not use old caches. In other words, cache keys must be different after policies are updated. That's why I added `PolicyPaths`, `DataPaths`, and `FilePatterns` for calculating cache keys.

knqyf263:
>hashes change intended? most of the updates on artifact/... tests are incorporating new hashes for blobs

I confirmed it was fine. It used to add analyzer versions as key suffixes like `sha256:beee9f30bc1f711043e78d4a2be0668955d4b761d587d6f60c2c8dc081efb203/110111`. But the cache keys would be new sha256 digests based on layer ID, analyzer versions, policy contents, data contents, and file patterns. It makes sense that keys are different.

cache/key_test.go
```diff
@@ -28,10 +32,107 @@ func TestWithVersionSuffix(t *testing.T) {
 			},
 			want: "sha256:1acfb4a0eb0acb0b1e8d462df18841581b960d5df190b32f3327df4820e5bc7b",
 		},
+		{
+			name: "empty slice file patterns",
+			args: args{
+				key:     "sha256:5c534be56eca62e756ef2ef51523feda0f19cd7c15bb0c015e3d6e3ae090bf6e",
+				version: "111101112110013",
+			},
+			patterns: []string{},
```
ronaudinho, line 41: confirmed to produce same result as unset or set as nil

cache/key_test.go
```diff
@@ -28,10 +32,107 @@ func TestWithVersionSuffix(t *testing.T) {
...
+		{
+			name: "non empty string in file patterns",
+			args: args{
+				key:     "sha256:5c534be56eca62e756ef2ef51523feda0f19cd7c15bb0c015e3d6e3ae090bf6e",
+				version: "111101112110013",
+			},
+			patterns: []string{"test"},
+			want:     "sha256:7bbca3f56a19cc7d892879d17ba2204f62d37fd9122ca74271e5bb4ab0817bc1",
+		},
+		{
```
ronaudinho, line 71: confirmed that empty strings in file patterns does not affect the hash

cache/key_test.go
```diff
@@ -28,10 +32,107 @@ func TestWithVersionSuffix(t *testing.T) {
...
+			name: "empty slice policy paths",
+			args: args{
+				key:     "sha256:5c534be56eca62e756ef2ef51523feda0f19cd7c15bb0c015e3d6e3ae090bf6e",
+				version: "111101112110013",
+			},
+			policy: []string{},
+			want:   "sha256:1acfb4a0eb0acb0b1e8d462df18841581b960d5df190b32f3327df4820e5bc7b",
+		},
+		{
```
ronaudinho, line 89: I have yet to find what []string{""} is evaluated too, but it probably depends on dir content too.

cache/key_test.go
```diff
@@ -28,10 +32,107 @@ func TestWithVersionSuffix(t *testing.T) {
...
+			name: "empty string in policy paths",
+			skip: "returns unique hash on each run",
+		},
+		{
+			name: "multiple empty string in policy paths",
+			skip: "returns unique hash on each run, different hash from single empty string",
+		},
+		{
```
ronaudinho, line 97: this is not testable as every time i change the expected hash, the resulting hash changes due to that said change.

anyway, all these skips are just for reference, should we omit these?

analyzer/config/hcl/testdata/deny.rego
ronaudinho: looks like because i copy pasted the initial example, and the indent is different from my editor settings. fixed.

cache/key_test.go
ronaudinho: updated

cache/key_test.go
knqyf263: Does "dot" mean the current directory? As you said, it is not testable. We don't need these cases.

cache/key_test.go
knqyf263: I'm guessing `[]string{""}` is the same as `[]string{"."}`.

cache/key_test.go
ronaudinho: yes "." as in current dir. okay will remove this then.

cache/key_test.go
ronaudinho: oh I misread and thought they produced different hashes, that's why I took note :facepalm: removing as well

ronaudinho:
added multiple rego rules test on `analyzer/config`:
- `docker`
    - multiple rules
    - single policy path at *analyzer/config/testdata/docker_multi.rego*
- `hcl`, `json`, `toml`, `yaml`
   - multiple rules
   - multiple policy paths at *analyzer/config/testdata/deny.rego* and *analyzer/config/testdata/warn.rego*

knqyf263 (approved):

knqyf263:
Thanks a lot!

PR #176, opened 2021-05-01
Skipping directory is file is symlink

rahul2393:
[Issue](https://github.com/aquasecurity/trivy/issues/968)


walker/fs.go
```diff
@@ -20,6 +20,8 @@ func WalkDir(root string, f WalkFunc) error {
 			return nil
 		} else if isIgnored(pathname) {
 			return filepath.SkipDir
+		} else if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
+			return filepath.SkipDir
```
knqyf263, line 24: Should return nil

knqyf263: And try IsRegular()

knqyf263: In that case, we no longer need IsDir()

knqyf263 (approved):

PR #181, opened 2021-05-27
feat: Support Google artifact registry

TauseefMalik:
This commit adds the capability to scan images from Google artifact
registry(GAR). GAR domains were earlier rejected by Trivy e.g.
europe-west3-docker.pkg.dev etc. With this change, we will treat domain
ending with 'docker.pkg.dev' as GAR domain and use gcloud sdk to fetch
credentials from provided file or credstore.

### DIFF
--- a/image/token/google/google.go
+++ b/image/token/google/google.go
@@ -1,4 +1,4 @@
-package gcr
+package google
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


knqyf263 (commented):
What is the difference from GCR? I mean why it is not enough to add a new suffix here.
https://github.com/aquasecurity/fanal/blob/main/image/token/gcr/gcr.go#L21

TauseefMalik:
Updated the PR

knqyf263 (approved):

PR #182, opened 2021-05-31
feat(image): add uncompressed layer size

tonaim:
This commit will help in getting uncompressed layer sizes. 
Can sum up these layer sizes to get the actual image size


knqyf263 (changes requested):
First of all, you need to fix tests. Also, UncompressedSize is expensive and performance got much worse. I replaced it with a custom reader.
https://github.com/google/go-containerregistry/blob/4c244d6cf55d8d1a4b7287b26651e99ae29b56cb/pkg/v1/partial/with.go#L327-L331

artifact/image/image.go
```diff
@@ -239,7 +239,9 @@ func (a Artifact) uncompressedLayer(diffID string) (string, io.Reader, error) {
 		}
 		digest = d.String()
 	}
-
+	if err != nil {
+		return "", nil, xerrors.Errorf("failed to get the uncompressed layer size (%s): %w", diffID, err)
+	}
```
knqyf263, line 244: cruft?

knqyf263 (approved):

PR #184, opened 2021-06-14
Custom cache support

ankk13:
Adding functionality to support custom cache and analyzer
Before inspection of the resources, missing layers will be collected from all the caches.
Union of all the missing layers will be analysed.
The analysis from each analyzer will be returned as map of cache and result.
Result blob will be updated in appropriate Cache

while getting blobs from cache while applying layers,
all the blobs for a layer will be pulled and combined blob will be refered
for merging the layer analysis.

The logic of merging layers for custom info is just for flow completion state.


types/artifact.go
```diff
@@ -82,24 +90,26 @@ type ArtifactInfo struct {
...
-	Packages          []Package          `json:",omitempty"`
-	Applications      []Application      `json:",omitempty"`
-	Misconfigurations []Misconfiguration `json:",omitempty"`
-	Size              int                `json:",omitempty"`
+	OS                *OS                          `json:",omitempty"`
+	Packages          []Package                    `json:",omitempty"`
+	Applications      []Application                `json:",omitempty"`
+	Misconfigurations []Misconfiguration           `json:",omitempty"`
+	Size              int                          `json:",omitempty"`
+	CustomResources   map[string]AnalyzerResources `json:",omitempty"`
```
knqyf263, line 112: `CustomResources` is unnecessary for ArtifactDetail now.

types/artifact.go
```diff
@@ -82,24 +90,26 @@ type ArtifactInfo struct {
...
+	Digest            string                       `json:",omitempty"`
+	DiffID            string                       `json:",omitempty"`
+	OS                *OS                          `json:",omitempty"`
+	PackageInfos      []PackageInfo                `json:",omitempty"`
+	Applications      []Application                `json:",omitempty"`
+	Misconfigurations []Misconfiguration           `json:",omitempty"`
+	OpaqueDirs        []string                     `json:",omitempty"`
+	WhiteoutFiles     []string                     `json:",omitempty"`
+	Size              int                          `json:",omitempty"`
+	CustomResources   map[string]AnalyzerResources `json:",omitempty"`
```
knqyf263, line 102: How about this?

```suggestion
	CustomResources   []CustomResource `json:",omitempty"`
```

We already have the `Type` field in CustomResource. So, we can marshal `Info` depending on the type. Then, we can merge these resources based on FilePath + Type.

```
		for _, customResource := range layer.CustomResources {	
			// e.g. key => /tmp/something/malicious/custom:malware
			key := fmt.Sprintf("%s/custom:%s", customResource.FilePath, customResource.Type)
			nestedMap.SetByString(key, sep, customResource)
		}
```

Also, we have to replace the keys of PackageInfos, Applications, and Configs.
https://github.com/aquasecurity/fanal/blob/048c06dd1202dabca54ac99207194d5732fe9bd6/applier/docker.go#L85-L97

Like
```
nestedMap.SetByString(fmt.Sprintf("%s/type:ospkg", pkgInfo.FilePath), sep, pkgInfo)
```

Then, you can write as below

```
	_ = nestedMap.Walk(func(keys []string, value interface{}) error {
		switch v := value.(type) {
		case types.PackageInfo:
			mergedLayer.Packages = append(mergedLayer.Packages, v.Packages...)
		case types.Application:
			mergedLayer.Applications = append(mergedLayer.Applications, v)
		case types.Config:
			mergedLayer.Configs = append(mergedLayer.Configs, v)
		case types.CustomResource:
			mergedLayer.CustomResources = append(mergedLayer.CustomResources, v)
		}
		return nil	})
```

types/artifact.go
```diff
@@ -52,6 +52,14 @@ type Application struct {
 	Libraries []LibraryInfo
 }
 
+type CustomResource struct {
+	FilePath string
+	Type     string
```
knqyf263, line 57: nit: be aligned with other types

```suggestion
	Type     string
	FilePath string
```

artifact/image/image.go
```diff
@@ -158,66 +206,76 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys []
...
+		if cache == types.BuiltInCache {
+			blobInfo.Misconfigurations = misconfs
+			blobInfo.OpaqueDirs = opqDirs
+			blobInfo.WhiteoutFiles = whFiles
+			blobInfo.OS = resultMap[cache].OS
+			blobInfo.PackageInfos = resultMap[cache].PackageInfos
+			blobInfo.Applications = resultMap[cache].Applications
+		} else {
+			blobInfo.CustomResources = resultMap[cache].CustomResources
+		}
```
knqyf263, line 275:
```suggestion
		blobInfo.Misconfigurations = misconfs
		blobInfo.OpaqueDirs = opqDirs
		blobInfo.WhiteoutFiles = whFiles
		blobInfo.OS = resultMap[cache].OS
		blobInfo.PackageInfos = resultMap[cache].PackageInfos
		blobInfo.Applications = resultMap[cache].Applications
		blobInfo.CustomResources = resultMap[cache].CustomResources
```

artifact/image/image.go
```diff
@@ -158,66 +206,76 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys []
...
-		return types.BlobInfo{}, xerrors.Errorf("walk error: %w", err)
+		return layerInfo, xerrors.Errorf("walk error: %w", err)
 	}
 
 	// Wait for all the goroutine to finish.
 	wg.Wait()
 
-	// Sort the analysis result for consistent results
-	result.Sort()
+	for cache, _ := range resultMap {
```
knqyf263, line 251: Any reason we cannot use the second argument?

```suggestion
	for cache, result := range resultMap {
```

artifact/image/image.go
```diff
@@ -158,66 +206,76 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys []
...
+			}
 		}
 	}
 
 	return nil
 
 }
 
-func (a Artifact) inspectLayer(ctx context.Context, diffID string) (types.BlobInfo, error) {
+func (a Artifact) inspectLayer(ctx context.Context, diffID string) (layerInfo map[types.CacheType]types.BlobInfo, err error) {
```
knqyf263, line 222:
```suggestion
func (a Artifact) inspectLayer(ctx context.Context, diffID string) (map[types.CacheType]types.BlobInfo, error) {
```

artifact/image/image.go
```diff
@@ -158,66 +206,76 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys []
...
 	log.Logger.Debugf("Missing diff ID: %s", diffID)
-
+	layerInfo = map[types.CacheType]types.BlobInfo{}
 	layerDigest, r, err := a.uncompressedLayer(diffID)
 	if err != nil {
-		return types.BlobInfo{}, xerrors.Errorf("unable to get uncompressed layer %s: %w", diffID, err)
+		return layerInfo, xerrors.Errorf("unable to get uncompressed layer %s: %w", diffID, err)
```
knqyf263, line 227:
```suggestion
		return nil, xerrors.Errorf("unable to get uncompressed layer %s: %w", diffID, err)
```

artifact/image/image.go
```diff
@@ -158,66 +206,76 @@ func (a Artifact) inspect(ctx context.Context, missingImage string, layerKeys []
...
 	opqDirs, whFiles, err := walker.WalkLayerTar(cr, func(filePath string, info os.FileInfo, opener analyzer.Opener) error {
-		if err = a.analyzer.AnalyzeFile(ctx, &wg, limit, result, filePath, info, opener); err != nil {
+		if err = a.analyzer.AnalyzeFile(ctx, &wg, limit, resultMap, filePath, info, opener); err != nil {
 			return xerrors.Errorf("failed to analyze %s: %w", filePath, err)
 		}
 		return nil
 	})
 	if err != nil {
-		return types.BlobInfo{}, xerrors.Errorf("walk error: %w", err)
+		return layerInfo, xerrors.Errorf("walk error: %w", err)
```
knqyf263, line 245:
```suggestion
		return nil, xerrors.Errorf("walk error: %w", err)
```

artifact/image/image.go
```diff
@@ -51,7 +51,30 @@ func NewArtifact(img image.Image, c cache.ArtifactCache, disabled []analyzer.Typ
...
 		image:               img,
-		cache:               c,
+		caches:              []cache.ArtifactCache{c},
+		analyzer:            analyzer.NewAnalyzer(disabled),
+		scanner:             s,
+		configScannerOption: opt,
+	}, nil
+}
+
+func NewArtifactMultiCache(img image.Image, c []cache.ArtifactCache, disabled []analyzer.Type, opt config.ScannerOption) (artifact.Artifact, error) {
```
knqyf263, line 61: We should update NewArtifact

applier/applier.go
```diff
@@ -8,32 +8,60 @@ import (
...
+
 func (a Applier) ApplyLayers(imageID string, diffIDs []string) (types.ArtifactDetail, error) {
 	var layers []types.BlobInfo
 	for _, diffID := range diffIDs {
-		layer, _ := a.cache.GetBlob(diffID)
-		if layer.SchemaVersion == 0 {
-			return types.ArtifactDetail{}, xerrors.Errorf("layer cache missing: %s", diffID)
+		// combining a1,b1,c1 = abc1
+		blobsInCache := map[types.CacheType]types.BlobInfo{}
+		for i := range a.cache {
```
knqyf263, line 41: Do not iterate cache here since ApplyLayers is supposed to be called for each cache respectively.

types/artifact.go
ankk13: Yes, map is not needed. I have replaced it with array.

types/artifact.go
ankk13: Yes, As the CustomResource schema is changed, Type can help us differentiate the custom resource. I have updated that follows this design.

knqyf263:
In favor of https://github.com/aquasecurity/fanal/pull/340

PR #208, opened 2021-07-13
Update tfsec to v0.46.0

liamg:
- Updated to latest tfsec (v0.46.0)
- tfsec now uses severities that pair with Trivy (low, medium, high, critical) so we were able to remove the conversion function
- Import paths for tfsec changed due to the move to `aquasecurity` organisation
- Adjusted tests to feature new severities, new checks, and extra info which is passed through for passed checks


knqyf263 (approved):

knqyf263:
Thanks!

PR #210, opened 2021-07-14
fix(image): layer size

tonaim:
while doing walk tar, we can get the actual size of layer from tar headers.


image/daemon/image.go
```diff
@@ -7,7 +7,7 @@ import (
 	"sync"
 
 	"github.com/docker/docker/api/types"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
knqyf263, line 10: needed?

knqyf263 (approved):

knqyf263:
Could you resolve conflicts?

image/daemon/image.go
tonaim: It is not needed. go imports added it. Anyhow removed alias now.

PR #217, opened 2021-07-28
feat(apk): capture license information

santhosh1729:
screenshot_32


analyzer/pkg/apk/apk.go
```diff
@@ -98,3 +104,8 @@ func (a alpinePkgAnalyzer) Type() analyzer.Type {
 func (a alpinePkgAnalyzer) Version() int {
 	return version
 }
+
+func replace(license string) string {
+	return strings.Replace(strings.Replace(license, " AND ", " ", -1), " OR ", " ", -1)
```
knqyf263, line 109: What does "AND" mean?

santhosh1729: If we have license information like this "MIT BSD AND GPL2+" , then replace space with comma then it becomes MIT,MIT,BSD,**AND**,GPL2+ like this. To avoid this, replacing the AND and OR values from license. 

FYI:  all package manager's license data which will be displayed with current license changes 

**apk:** MIT BSD AND GPL2+ => if replace space with comma then it becomes MIT,MIT,BSD,AND,GPL2+
**rpm :** LGPLv2+ and LGPLv2+ with exceptions and GPLv2+ and GPLv2+ with exceptions and BSD and Inner-Net and ISC and Public Domain and GFDL (showing as it is. Not removing AND, OR values)
**dpkg:** GAP,GPL,GPL-2,GPL-2+,LGPL,LGPL-2,LGPL-2+,LGPL-2.1+,other,public domain,public-domain (adding all license info with comma separator)

knqyf263 (approved):

PR #218, opened 2021-07-28
feat(rpm): capture license

santhosh1729:
screenshot_33


analyzer/pkg/rpm/rpm.go
```diff
@@ -95,6 +95,7 @@ func (a rpmPkgAnalyzer) parsePkgInfo(packageBytes []byte) (pkgs []types.Package,
 			SrcVersion:      srcVer,
 			SrcRelease:      srcRel,
 			Modularitylabel: pkg.Modularitylabel,
+			License:         strings.Replace(strings.Replace(pkg.License, " and ", ",", -1), " or ", ",", -1),
```
knqyf263, line 98: I don't think "and" and "or" are the same. Is there any specification on the license field in RPM DB? If no, we should keep this field as is.

knqyf263 (approved):

knqyf263:
I didn't look into tests, but it is failing now.

PR #224, opened 2021-08-04
Wordpress detection

ankk13:
Composer parser is modified to consider 2 types of file.
1. composer.lock
2. /wp-includes/version.php
This new file helps us to detect version of wordpress.

### DIFF
--- a/analyzer/library/composer/composer.go
+++ b/analyzer/library/composer/composer.go
@@ -1,16 +1,17 @@
 package composer
 
 import (
+	"bytes"
 	"os"
-	"path/filepath"
+	"strings"
 
 	"golang.org/x/xerrors"
 
 	"github.com/aquasecurity/fanal/analyzer"
 	"github.com/aquasecurity/fanal/analyzer/library"
 	"github.com/aquasecurity/fanal/types"
-	"github.com/aquasecurity/fanal/utils"
 	"github.com/aquasecurity/go-dep-parser/pkg/composer"
+	godepTypes "github.com/aquasecurity/go-dep-parser/pkg/types"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


knqyf263 (commented):
It should be outside of the composer analyzer.

analyzer/library/composer/composer.go
```diff
@@ -19,21 +20,38 @@ func init() {
...
-		return nil, xerrors.Errorf("error with composer.lock: %w", err)
+	parsedLibs := []godepTypes.Library{}
+	var err error
+	r := bytes.NewReader(target.Content)
+	if strings.HasSuffix(target.FilePath, "/composer.lock") {
+		parsedLibs, err = composer.Parse(r)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to parse %s: %w", target.FilePath, err)
+		}
+	} else if strings.HasSuffix(target.FilePath, "/wp-includes/version.php") {
```
masahiro331, line 36: IMO: If you want to use meaningful strings in multiple places, it is preferable to define them const.
like.
```
const (
    composer = "/composer.lock"
    wordpress = "/wp-includes/version.php"
)
```

analyzer/library/composer/composer.go
```diff
@@ -19,21 +20,38 @@ func init() {
...
+		parsedLibs, err = composer.ParseWordPress(r)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to parse %s: %w", target.FilePath, err)
+		}
 	}
-	return res, nil
+	if len(parsedLibs) == 0 {
```
masahiro331, line 42: Other libraries return an AnalysisResult even if the detected library is empty.

analyzer/library/composer/composer_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+
+	"github.com/aquasecurity/fanal/analyzer"
+	"github.com/aquasecurity/fanal/types"
+	godeptypes "github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_composer_Analyze(t *testing.T) {
```
masahiro331, line 14: Please add more invalid case tests.

analyzer/library/composer/composer.go
masahiro331: Sorry, no problems skip this process. 
https://github.com/aquasecurity/fanal/blob/2cd043a8486e276a20c5607814e2972e0cbb3db4/analyzer/analyzer.go#L98-L100

knqyf263:
We decided not to implement WordPress now.

PR #227, opened 2021-08-08
feat(image): add image metadata

tonaim:
made below changes.
1. As layer size calculation is not correct. Removed the code related to it.
2. Added changes to get artifact metadata from inspect method.


image/daemon/image.go
```diff
@@ -92,8 +98,45 @@ func (img *image) ConfigFile() (*v1.ConfigFile, error) {
...
+	loc, err := time.LoadLocation("")
+	if err != nil {
+		return nil, xerrors.Errorf("failed setting location %s: %w", img.inspect.Created, err)
+	}
+	time.Local = loc
+	imgCreatedDate, err := dateparse.ParseLocal(img.inspect.Created)
+	fmt.Println(imgCreatedDate)
+	if err != nil {
+		return nil, xerrors.Errorf("failed parsing created Date %s: %w", img.inspect.Created, err)
+	}
```
knqyf263, line 131: Could you elaborate on it? Doesn't `time.ParseLocation(img.inspect.Created, time.UTC)` work?

types/artifact.go
```diff
@@ -111,8 +115,15 @@ type ArtifactDetail struct {
...
 	// HistoryPackages are packages extracted from RUN instructions
 	HistoryPackages []Package `json:",omitempty"`
 }
+
+type LayerMetadata struct {
+	Digest           string     `json:",omitempty"`
+	LayerDiffId      string     `json:",omitempty"`
+	UncompressedSize int64      `json:",omitempty"`
+	LayerHistory     v1.History `json:",omitempty"`
+}
```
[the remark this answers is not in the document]
tonaim: I agree history is not needed here. But these are not saved in cache. 

1. From inspect (client) we get v1.config (which will have all metadata like architecture, docker vesion etc). But will not have layer digest
2. So the digest will get it from scan (server) to complete the metadata.  I left the size place holder as it is to use it in future.

image/daemon/image.go
tonaim: Sure, I assume you are asking about `ParseInLocation`  in https://golang.org/src/time/format.go.

It needs layout as input, that is the problem here. We don't know the date layout of created date.

knqyf263: Why don't we know the date layout?

types/artifact.go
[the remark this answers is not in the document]
knqyf263: `Digest` cannot be obtained when scanning local images even in the server.

knqyf263 (approved):

PR #209, opened 2021-07-14
Update dependencies and fix security issue

mmorel-35:
This also includes dependabot to handle security update for gomod and github-actions

PR #228, opened 2021-08-09
chore(ci) : Setup dependabot and fix security issues

mmorel-35:
Replaces #209
Concerning gomod updates, I know that you are planning to do it by trivy but as it not active yet, I'm suggesting to do it with Dependabot first.


.github/workflows/test.yml
```diff
@@ -16,7 +16,13 @@ jobs:
...
 
     - name: Check out code into the Go module directory
-      uses: actions/checkout@v1 
+      uses: actions/checkout@v1
+    - uses: actions/cache@v2
+      with:
+        path: ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
```
knqyf263, line 25: `Install dependencies` takes only 10 seconds. I don't think we need to use cache now.
image

knqyf263 (commented):
We are discussing tfsec update in https://github.com/aquasecurity/fanal/pull/229. Please do not update tfsec yet.

.github/workflows/test.yml
mmorel-35: Done :) !

knqyf263 (dismissed):

mmorel-35:
Please relaunch the Job , this test passes once one time on two

mmorel-35:
It doesn't work on https://github.com/mmorel-35/fanal/actions/runs/1119965762
But it passes on https://github.com/mmorel-35/fanal/actions/runs/1119969448 and I haven't touched this part

knqyf263 (approved):

knqyf263 (approved):

knqyf263:
Thanks

knqyf263:
I couldn't find how to increment only the major version of GitHub Actions. @mmorel-35 Do you know some tips? If no, I'll revert Dependabot.
https://github.com/aquasecurity/fanal/pull/232

PR #229, opened 2021-08-10
Update tfsec to v0.57.1

liamg:


knqyf263 (commented):
For consistency, I want to keep a numeric ID since all other vulnerabilities have CVE-ID or something similar such as CVE-2021-1234 and RHSA-2021-5678. Also, other IaC scanning has a similar ID like KSV023. We should not use descriptive ID only for tfsec. Is it possible to have both descriptive and numeric IDs in tfsec?

liamg:
Our users raised the issue that inline ignore rules were hard to make sense of as AWS001, AWS003, AWS097 etc. simply don't convey as much meaning as the more verbose IDs, which is why we changed them. There are currently no alternative numeric IDs to use, though we are talking about the possibility of assigning each check an "Aqua ID" to link to the AVD so perhaps we could use that in future?

knqyf263:
Yes, we can probably use Aqua ID.

liamg:
Will reopen with a future version once we have nicer IDs :+1:

knqyf263:
We'll look forward to it!

PR #254, opened 2021-08-22
feat: add hooks

knqyf263:
Hooks are called after analysis so that separate results from each analyzer can be aggregated or processed together.


ankk13 (approved):
Looks good to me.

tonaim (approved):

PR #262, opened 2021-09-08
fix(oci): handling of complex image indexes

afdesk:
Add a handling of image indexes that contain links to another image manifests.

Fixes [aquasecurity/trivy#1080](https://github.com/aquasecurity/trivy/issues/1080)


knqyf263 (changes requested):
Great! Thanks! You may forget applying `go fmt` though.

afdesk:
@knqyf263 this PR isn't final, because there isn't a way to select one of inner images

knqyf263 (approved):

knqyf263:
It is enough to take the first image for now.

PR #285, opened 2021-09-24
fix(tar): skip large non executable files except go binary files insi…

afdesk:
…de tar archive

If the file size inside tar archive is more than `TRIVY_LARGE_FILE_SIZE`, it'll skip this one, expect go binary files.

Fixes [aquasecurity/trivy#1242](https://github.com/aquasecurity/trivy/issues/1242)

### DIFF
--- a/walker/tar.go
+++ b/walker/tar.go
@@ -2,9 +2,12 @@ package walker
 
 import (
 	"archive/tar"
+	"debug/elf"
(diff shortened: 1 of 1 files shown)


knqyf263:
Close this PR in favor of https://github.com/aquasecurity/fanal/pull/314

PR #291, opened 2021-09-28
Chore(dep): update the tfsec version

owenrumney:
- Latest tfsec version has many more checks and also includes the AVD ID


liamg (approved):

config/scanner/scanner_test.go
```diff
@@ -72,40 +72,40 @@ func TestScanner_ScanConfig(t *testing.T) {
...
 							Severity:           "CRITICAL",
+							RecommendedActions: "Don't include sensitive data in blocks",
 						},
 					},
 					{
 						Message: `Resource 'azurerm_managed_disk.source' passed check: Potentially sensitive data stored in block attribute.`,
 						PolicyMetadata: types.PolicyMetadata{
 							Type:               "Terraform Security Check powered by tfsec",
-							ID:                 "GEN003",
+							ID:                 "AVD-GEN-0001",
```
knqyf263, line 108: I couldn't find this ID in AVD. Is it planned to be added?
https://avd.aquasec.com/

owenrumney: It's here - https://github.com/aquasecurity/cloud-metadata/blob/main/general/secrets/sensitive-in-attribute/metadata.json 

The idea of this repo is that all IAC and CSPM rules will share the same set of common metadata but have different remediation steps depending on the reason (cli to resolve issues already in production maybe, or the required changes to terraform or cloudformation to fix the issue in code)

knqyf263: Do you mean AVD will not have the ID? Should Trivy link these IDs to TFsec doc directly? I thought all information would be aggregated in AVD and Trivy should link IDs to AVD.

knqyf263: OK, thanks!

knqyf263: Trivy currently refers to the tfsec documentation (tfsec.dev). In other words, this change would not be a regression in terms of references. I think we can merge this PR now and will switch references to AVD once AVD will be ready.

knqyf263 (approved):

knqyf263:
Seems like quay.io is down now. I will retry it later.

knqyf263:
Thanks!

PR #294, opened 2021-09-30
fix(go-binary): skip large files

afdesk:
Add skipping large files whose size is more than 200MB.

Fixes [aquasecurity/trivy#1242](https://github.com/aquasecurity/trivy/issues/1242)

### DIFF
--- a/analyzer/language/golang/binary/binary.go
+++ b/analyzer/language/golang/binary/binary.go
@@ -36,6 +36,11 @@ func (a gobinaryLibraryAnalyzer) Required(_ string, fileInfo os.FileInfo) bool {
 		return false
 	}
 
+	const MaxFileSize = 200000000
+	if fileInfo.Size() > MaxFileSize {
+		return false
+	}
+
 	// Check executable file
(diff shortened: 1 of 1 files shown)


analyzer/language/golang/binary/binary.go
```diff
@@ -36,6 +36,11 @@ func (a gobinaryLibraryAnalyzer) Required(_ string, fileInfo os.FileInfo) bool {
 		return false
 	}
 
+	const MaxFileSize = 200000000
```
knqyf263, line 39: We should define it outside of the function. Also, we can change it human-friendly.
```suggestion
	const (
		MiB = 1024*1024
		MaxFileSize = 200 * MiB
	)
```

knqyf263 (changes requested):

analyzer/language/golang/binary/binary.go
afdesk: changed.

knqyf263 (approved):

PR #295, opened 2021-10-01
Add ctx object to analyser to get values stored in ctx

santhosh1729:
for custom javascript/java analysers, aqua stores feeds in ctx, To process javascript/java packages, we need to pass ctx object to analysers

### DIFF
--- a/analyzer/analyzer.go
+++ b/analyzer/analyzer.go
@@ -31,6 +31,7 @@ type AnalysisTarget struct {
 	Dir      string
 	FilePath string
 	Content  []byte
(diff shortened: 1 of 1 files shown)


analyzer/analyzer.go
```diff
@@ -31,6 +31,7 @@ type AnalysisTarget struct {
 	Dir      string
 	FilePath string
 	Content  []byte
+	Ctx      context.Context
```
knqyf263, line 34: It is not relevant to the analysis target. It should be outside.

PR #296, opened 2021-10-01
fix(config/yaml): support yaml files which incompatible with json spec

mozillazg:
https://github.com/aquasecurity/trivy/issues/1251
https://github.com/aquasecurity/trivy/issues/1183

root cause: https://github.com/go-yaml/yaml/issues/139

### DIFF
--- a/analyzer/config/yaml/testdata/incompatible_json.yaml
+++ b/analyzer/config/yaml/testdata/incompatible_json.yaml
@@ -1,0 +1,4 @@
+replacements:
+  amd64: 64bit
+  386: 32bit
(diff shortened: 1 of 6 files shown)


knqyf263:
Does it mean sigs.k8s.io unmarshals into map[string]interface{}?

mozillazg:
@knqyf263 Yes, it'll try to make the result of Unmarshal be compatible with JSON

knqyf263 (approved):

knqyf263:
Thanks for the great contribution!

PR #314, opened 2021-10-23
fix(analyzer): improve performance

masahiro331:
## Fix the following issue  
https://github.com/aquasecurity/trivy/issues/1295  
https://github.com/aquasecurity/trivy/issues/1284

## Description 

This pull request modifies the walker implementation to improve the amount of memory with some analyzers.

**For Filesystem  command case**

If the file size is large
1. Run os.Open for each execution
2. Also return the clean function and delete the file after everything is done

If the file size is small, read the file only once

**For Image command case**

When the file size is large
1. Get the file from tar and write it to a local file only once
2. Cache the temp file name
3. Open for each execution. 
4. Delete the file after everything is done

If the file size is small, read the file only once


analyzer/language/python/packaging/packaging.go
```diff
@@ -43,18 +44,22 @@ type packagingAnalyzer struct{}
 
 // Analyze analyzes egg and wheel files.
 func (a packagingAnalyzer) Analyze(_ context.Context, target analyzer.AnalysisTarget) (*analyzer.AnalysisResult, error) {
-	content := target.Content
+	var r = target.ContentReader
 
 	// .egg file is zip format and PKG-INFO needs to be extracted from the zip file.
 	if strings.HasSuffix(target.FilePath, ".egg") {
+		content, err := ioutil.ReadAll(r)
```
knqyf263, line 51: https://go.dev/doc/go1.16#ioutil

analyzer/language/python/packaging/packaging.go
```diff
@@ -43,18 +44,22 @@ type packagingAnalyzer struct{}
...
+		if err != nil {
+			return nil, xerrors.Errorf("failed to read the egg file: %w", err)
+		}
+
 		pkginfoInZip, err := a.analyzeEggZip(content)
 		if err != nil {
 			return nil, xerrors.Errorf("egg analysis error: %w", err)
 		}
-		content = pkginfoInZip
+		r = bytes.NewReader(pkginfoInZip)
```
knqyf263, line 60: analyzeEggZip should return io.Reader

analyzer/language/analyze_test.go
```diff
@@ -29,9 +30,9 @@ func TestAnalyze(t *testing.T) {
 		{
 			name: "happy path",
 			args: args{
-				analyzerType: types.GoBinary,
-				filePath:     "app/myweb",
-				content:      []byte("happy"),
+				analyzerType:  types.GoBinary,
+				filePath:      "app/myweb",
+				contentReader: bytes.NewReader([]byte("happy")),
```
knqyf263, line 35: We may want to use strings.NewReader

analyzer/config/yaml/yaml.go
```diff
@@ -30,11 +31,16 @@ func NewConfigAnalyzer(filePattern *regexp.Regexp) ConfigAnalyzer {
 }
 
 func (a ConfigAnalyzer) Analyze(_ context.Context, target analyzer.AnalysisTarget) (*analyzer.AnalysisResult, error) {
+	content, err := ioutil.ReadAll(target.ContentReader)
```
knqyf263, line 34: https://go.dev/doc/go1.16#ioutil

analyzer/config/hcl/hcl.go
```diff
@@ -52,13 +53,18 @@ func (a ConfigAnalyzer) analyze(target analyzer.AnalysisTarget) (interface{}, er
 	var errs error
 	var parsed interface{}
 
-	if err := hcl2.Unmarshal(target.Content, &parsed); err != nil {
+	content, err := ioutil.ReadAll(target.ContentReader)
```
knqyf263, line 56: ditto

walker/tar.go
```diff
@@ -91,3 +95,67 @@ func underSkippedDir(filePath string, skipDirs []string) bool {
 	}
 	return false
 }
+
+// fileWithTarOpener opens a file in a Tar.
+// If the file size is greater than or equal to ThresholdSize(200MB), it writes the file and caches the file name.
+// If the file size is less than ThresholdSize(200MB), it opens the file once and the content is shared so that some analyzers can use the same data
+func (w *walker) fileWithTarOpener(fi os.FileInfo, r io.Reader) func() (io.ReadCloser, func() error, error) {
```
knqyf263, line 102: We can simply pass the file size instead of os.FileInfo.

walker/tar.go
```diff
@@ -91,3 +95,67 @@ func underSkippedDir(filePath string, skipDirs []string) bool {
...
+
+	return func() (io.ReadCloser, func() error, error) {
+		once.Do(func() {
+			if fi.Size() >= ThresholdSize {
+				var f *os.File
+				tempDirPath, err = ioutil.TempDir("", "trivy-*")
+				if err != nil {
+					err = xerrors.Errorf("failed to create the temp dir: %w", err)
+					return
+				}
```
knqyf263, line 118: Any reason to create a temp directory?  Looks like `os.CreateTemp` is enough here.

walker/walk.go
```diff
@@ -18,6 +14,8 @@ var (
 	systemDirs = []string{"proc", "sys", "dev"}
 )
 
+const ThresholdSize = int64(200) << 20
```
knqyf263, line 17:
```suggestion
const ThresholdSize = int64(200) << 20 // 200MB
```

analyzer/analyzer.go
```diff
@@ -195,10 +194,16 @@ func (a Analyzer) AnalyzeFile(ctx context.Context, wg *sync.WaitGroup, limit *se
...
+		rc, cleaner, err := opener()
 		if err != nil {
 			return xerrors.Errorf("unable to open a file (%s): %w", filePath, err)
 		}
+		defer func() {
+			err := cleaner()
+			if err != nil {
+				log.Logger.Warn("Clean temp directory error: %s", err)
+			}
+		}()
```
knqyf263, line 206: Seems like this defer will be called many times for the same file and it causes an error possibly. We have to be careful about when we call `cleaner()`.

analyzer/analyzer.go
```diff
@@ -208,15 +213,18 @@ func (a Analyzer) AnalyzeFile(ctx context.Context, wg *sync.WaitGroup, limit *se
...
 			ret, err := a.Analyze(ctx, target)
 			if err != nil && !xerrors.Is(err, aos.AnalyzeOSError) {
 				log.Logger.Debugf("Analysis error: %s", err)
 				return
 			}
 			result.Merge(ret)
-		}(d, AnalysisTarget{Dir: dir, FilePath: filePath, Content: b})
+		}(d, AnalysisTarget{Dir: dir, FilePath: filePath, ContentReader: rc})
 	}
+	wg.Wait()
```
knqyf263, line 226: What made you add this here? It looks problematic.

knqyf263 (changes requested):

AkselAllas:
This seems like an important improvement. Any timeline for this?

knqyf263 (approved):

Issue #323, opened 2021-11-05
requirements.txt line continuation causes "python version error"


kohtala:
## Description

Testing with `trivy fs .` gives warnings for requirements.txt and fails to find Python vulnerabilities. I generate requirements.txt using `pip-compile --generate-hashes` from [pip-tools](https://github.com/jazzband/pip-tools) 6.4.0.

This produces requirements.txt with entries like

```
thinc==7.4.5 \
    --hash=sha256:02b71ae5a0fa906a0aca968bd65589e0ab9fabd511e57be839774228b1509224 \
    --hash=sha256:10bafe5ddce698180098345b9c55f762dc3456558be844d35d64175e511581b6 \
    --hash=sha256:24086aa0fb72f466782115d529574a825c89afa62eb817962b9339f61ab50e0d \
    --hash=sha256:29a47ad0289dda0520b5af8538b30e8134553130200b83c34311feb71739968d \
    --hash=sha256:309ec4cae81f4de2e4e4fbd0bcb52b10bef4b1a6352c6a9143f6a53d3b1060ef \
    --hash=sha256:5743fde41706252ec6ce4737c68d3505f7e1cc3d4431174a17149838d594f8cb \
    --hash=sha256:5774007b5c52501cab5e2970cadca84923b4c420fff06172f2d0c86531973ce8 \
    --hash=sha256:8b647de79fe5f98cd327983bf0e27d006b48ad9694ceabdb9a3832b614ed1618 \
    --hash=sha256:c408ab24b24e6368ce4b6ddebb579118042a22d3f2f2c4e19ca67e3eadc9ed33 \
    --hash=sha256:c43ed753aa70bc619e42e168be4926c8a47799af6121ff0727ba99b330afbb44 \
    --hash=sha256:cce68c5ea54cd32cef661858363509afdedad047027e8cdf0dc4edec0c2cc010 \
    --hash=sha256:d01ab1480d37ebefcac22d63ffe01916c9f025ae3dbdbe5824ac3ea5cce8e3fd \
    --hash=sha256:fae320de65af70786c1526ffc33b88f2da650d3106f5f9a06b37f0ac3944a44f
    # via spacy
```

## What did you expect to happen?

It reports the vulnerability "pyup.io-40660" suggesting upgrade to "8.0.4".

## What happened instead?

```
2021-11-05T11:13:28.944+0200    WARN    python version error (7.4.5\): malformed version: 7.4.5\

requirements.txt (pip)
==================================
[...]

PR #330, opened 2021-11-11
chore: update go-dep-parser dependency to fix pip parsing issue

rahul2393:
Updating go-dep-parser dependency to include fix for issue #323


knqyf263 (approved):
LGTM👍🏻

PR #331, opened 2021-11-18
feat: Secondary cache support

ankk13:
Added extra(secondary) caches support for artifact, which allows connecting additional cache for specific analyzers.
All analyzers need to implement a new method CacheType(). Existing analyzers will return build-in cache as cachetype. 
All caches should implement Type() method. Built-in cache is primary cache.


analyzer/command/apk/apk.go
```diff
@@ -59,6 +59,10 @@ type pkg struct {
 
 type version map[string]int
 
+func (a alpineCmdAnalyzer) CacheType() types.CacheType {
+	return types.BuiltInCache
+}
+
```
knqyf263, line 65: Embedding struct looks better since all built-in analyzers should have the same behavior.
https://go.dev/doc/effective_go#embedding

knqyf263:
In favor of https://github.com/aquasecurity/fanal/pull/340

PR #332, opened 2021-11-24
fix(hooks): exclude go from filtering

AndreyLevchenko:


hook/filter/filter.go
```diff
@@ -22,6 +22,23 @@ var (
...
+		//python
+		types.Pip,
+		types.Pipenv,
+		types.Poetry,
+		types.PythonPkg,
+
+		//node.js
+		types.NodePkg,
+		types.Npm,
+		types.Yarn,
```
knqyf263, line 40: The target is individual packages, not lock files.

```suggestion
		// ruby
		types.GemSpec,

		// python
		types.PythonPkg,

		// node.js
		types.NodePkg,
```

knqyf263 (approved):

PR #333, opened 2021-11-24
fix(applier): fixed layer applications update

DmitriyLewen:
Closes  [trivy issue #1407](https://github.com/aquasecurity/trivy/issues/1407)


knqyf263 (approved):

knqyf263:
Thanks!

Issue #334, opened 2021-11-27
Feature: Allow to suppress output of git clone


a-zen:
# Description

I was looking into the trivy issue [927](https://github.com/aquasecurity/trivy/issues/927) and noticed that the fanal library needs to be changed to allow to suppress the git clone output.

# What did you expect to happen?
The git _NewArtifact_ should have an option to suppress the git clone output

knqyf263:
#335 got merged

PR #335, opened 2021-11-27
allow suppressing the git clone output by setting the quiet option

a-zen:
Hi,

this is my proposal to support #334 which is needed to fix trivy issue [928](https://ghttps://github.com/aquasecurity/trivy/issues/927ithub.com/aquasecurity/trivy/issues/927).

Given the current architecture I found it easiest to extend the artifact options for this use case even tough the option probably doesn't fit a 100 % in there.

Any feedback is much appreciated.


knqyf263 (approved):

knqyf263:
Thanks a lot!!

PR #341, opened 2021-12-10
Also detect RPM databases in NDB format (Fixes #183)

### LINKED ISSUE #183: missing support for rpm-nbd
registry.suse.com/suse/sle15:15.3 and likely other containers are using rpm not with berkeleydb as backend but using the ndb backend for licensing reasons. (ndb is new functionality in rpm 4.14+). 

the current fanal code can not handle this as it is looking for a "/var/lib/rpm/Packages" file, which does not exist. in ndb the file is called `/var/lib/rpm/Packages.db` and the fileformat is different, so the existing parser is failing. 

previously this worked via the rpmcmd fallback, however that was removed in https://github.com/aquasecurity/fanal/pull/142

any recommendation on how to recover cointainer checking?

dirkmueller:
RPM databases in the native DB format rather than the traditional
Berkeley DB format are stored as different filenames, so detect
those as well.

(This needs a rebase on top of a newer version of go-rpmdb when https://github.com/knqyf263/go-rpmdb/pull/9 is merged)

Fix https://github.com/aquasecurity/fanal/issues/183

### DIFF
--- a/analyzer/pkg/rpm/rpm.go
+++ b/analyzer/pkg/rpm/rpm.go
@@ -23,7 +23,9 @@ func init() {
 const version = 1
 
 var requiredFiles = []string{
+	"usr/lib/sysimage/rpm/Packages.db",
 	"usr/lib/sysimage/rpm/Packages",
+	"var/lib/rpm/Packages.db",
 	"var/lib/rpm/Packages",
 }
 var errUnexpectedNameFormat = xerrors.New("unexpected name format")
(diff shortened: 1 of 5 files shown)


knqyf263:
Hi @dirkmueller, thanks for your contribution. We have sent our Corporate CLA to your email. Could you sign it and send it back to us?

dirkmueller:
requires publishing the image from https://github.com/aquasecurity/trivy-test-images/pull/6 in the docker registry organisation.

dirkmueller:
@knqyf263 I'm working with corp legal team to get it signed back to you.

knqyf263 (approved):
Thanks a lot!!

PR #342, opened 2021-12-15
fix: fixed skipFiles/skipDirs flags for relative path

DmitriyLewen:
Closes [trivy issue1479](https://github.com/aquasecurity/trivy/issues/1479)


knqyf263 (approved):

knqyf263:
Thanks!

PR #343, opened 2021-12-20
feat: added skip-cache flag for images

DmitriyLewen:
added boolean flag `skip-cache` to skip cache when scanning images.

### DIFF
--- a/artifact/artifact.go
+++ b/artifact/artifact.go
@@ -14,6 +14,7 @@ type Option struct {
 	DisabledHooks     []hook.Type
 	SkipFiles         []string
 	SkipDirs          []string
+	SkipCache         bool
 	Quiet             bool
 }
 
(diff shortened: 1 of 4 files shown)


knqyf263:
What motivates you to add this feature?

AndreyLevchenko:
> What motivates you to add this feature?

Hi @knqyf263 
I was going to ask you about it. I'm 100% agree it has no value for customers. But we would love to use it while debuging issues. It would exclude extra step for removing fanal cache. If I miss this step accidentally it can lead investigations to a wrong path.

knqyf263:
$ trivy image --clear-cache && trivy image [YOUR_COMMAND]
doesn't work for you? I think we might be missing `--skip-cache` as well.

AndreyLevchenko:
I suspect it won't work in some debug configs. It could be something other then cmd option but it would be nice to have cache bypass logic

knqyf263:
>I suspect it won't work in some debug configs.

Could you give me some examples? I still don't get your point. `--clear-cache` has been working for me.

PR #349, opened 2021-12-23
feat(java): support offline mode

knqyf263:
## Description
This option allows avoiding Internet access. The results with/without the option may differ. For example, the offline mode will not try to resolve transitive dependencies in pom.xml when the dependency doesn't exist in the local repositories. On the other hand, the online mode will resolve those dependencies in the remote repositories. It means the dependencies might be less in offline mode.

## Supported Analyzers
- Java
  - [x] pom.xml
  - [x] JAR

## Blockers
- [x] https://github.com/aquasecurity/go-dep-parser/pull/18
- [x] https://github.com/aquasecurity/go-dep-parser/pull/70

### DIFF
--- a/analyzer/language/nodejs/npm/npm.go
+++ b/analyzer/language/nodejs/npm/npm.go
@@ -24,7 +24,7 @@ var requiredFiles = []string{"package-lock.json"}
 
 type npmLibraryAnalyzer struct{}
 
-func (a npmLibraryAnalyzer) Analyze(ctx context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
+func (a npmLibraryAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
 	res, err := language.Analyze(types.Npm, input.FilePath, input.Content, npm.Parse)
 	if err != nil {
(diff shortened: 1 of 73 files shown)


analyzer/language/nodejs/npm/npm.go
```diff
@@ -24,8 +24,8 @@ var requiredFiles = []string{"package-lock.json"}
 
 type npmLibraryAnalyzer struct{}
 
-func (a npmLibraryAnalyzer) Analyze(_ context.Context, target analyzer.AnalysisTarget) (*analyzer.AnalysisResult, error) {
-	res, err := language.Analyze(types.Npm, target.FilePath, target.Content, npm.Parse)
+func (a npmLibraryAnalyzer) Analyze(ctx context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
```
AndreyLevchenko, line 27: `_ context.Context` explains context usage better

AndreyLevchenko (changes requested):

analyzer/language/nodejs/npm/npm.go
knqyf263: Fixed https://github.com/aquasecurity/fanal/pull/349/commits/d1a3828b31baa21a3fea7bb957292d6282270f35

AndreyLevchenko (approved):

PR #354, opened 2022-01-05
fix: crash of handling compressed layers.

afdesk:
Often, the "compressed" bytes are not actually gzip-compressed.
It fixed commit [google/go-containerregistry@a65b7844a475e5bf08874dea33eb9529020834eb](https://github.com/google/go-containerregistry/commit/a65b7844a475e5bf08874dea33eb9529020834eb)

Fixes [aquasecurity/trivy#1527](https://github.com/aquasecurity/trivy/issues/1527)


knqyf263 (approved):
:shipit:

PR #355, opened 2022-01-06
Allow more filenames for the PIP requirements files

jerbob92:
This PR extends the python/pip analyzer to allow more flexible filenames.
It's quite common to split up the requirements.txt file into multiple files like requirements-dev.txt and requirements-prod.txt.

Dependabot also looks for files like this.


jerbob92:
@knqyf263 would you also want this for Dockerfile? I often see/use Dockerfile-dev and Dockerfile-prod for example.

PR #356, opened 2022-01-06
Allow to scan a single file

jerbob92:
This PR adds support to scan a single file.

It didn't need much to make this work, only difference is that when you have a filepath like this:

`/home/jeroen/test.txt` and you would pass that into fanal, it would try to make `/home/jeroen/test.txt` relative to `/home/jeroen/test.txt` (because it expects a folder as input), which results in the filename `.`, which is not correct.

I added some logic to extract the actual directory from the filePath when a full file path is given as input in stead of a directory (in that case `a.dir` and `filePath` are equal).


artifact/local/fs.go
```diff
@@ -82,14 +82,22 @@ func (a Artifact) Inspect(ctx context.Context) (types.ArtifactReference, error)
 	limit := semaphore.NewWeighted(parallel)
 
 	err := a.walker.Walk(a.dir, func(filePath string, info os.FileInfo, opener analyzer.Opener) error {
+		directory := a.dir
+
+		// When the directory is the same as the filePath, a file was given
+		// instead of a directory, rewrite the directory in this case.
+		if a.dir == filePath {
+			directory = filepath.Dir(a.dir)
```
knqyf263, line 90: We may need `filepath.Clean` here or somewhere because the leading `./` fails the scanning.

```
$ trivy fs --severity HIGH go.sum
2022-01-10T17:32:25.881+0200    INFO    Number of language-specific files: 1
2022-01-10T17:32:25.881+0200    INFO    Detecting gomod vulnerabilities...

go.sum (gomod)
==============
Total: 3 (HIGH: 3)

+-----------------------------+------------------+----------+--------------------+-------------------------+---------------------------------------+
|           LIBRARY           | VULNERABILITY ID | SEVERITY | INSTALLED VERSION  |      FIXED VERSION      |                 TITLE                 |
+-----------------------------+------------------+----------+--------------------+-------------------------+---------------------------------------+
| github.com/apache/thrift    | CVE-2020-13949   | HIGH     | 0.13.0             | v0.14.0                 | libthrift: potential DoS when         |
|                             |                  |          |                    |                         | processing untrusted payloads         |
|                             |                  |          |                    |                         | -->avd.aquasec.com/nvd/cve-2020-13949 |
+-----------------------------+------------------+          +--------------------+-------------------------+---------------------------------------+
| github.com/dgrijalva/jwt-go | CVE-2020-26160   |          | 3.2.0+incompatible |                         | jwt-go: access restriction            |
|                             |                  |          |                    |                         | bypass vulnerability                  |
|                             |                  |          |                    |                         | -->avd.aquasec.com/nvd/cve-2020-26160 |
+-----------------------------+------------------+          +--------------------+-------------------------+---------------------------------------+
| k8s.io/kubernetes           | CVE-2020-8558    |          | 1.13.0             | 1.16.11, 1.17.7, 1.18.4 | kubernetes: node localhost services   |
|                             |                  |          |                    |                         | reachable via martian packets         |
|                             |                  |          |                    |                         | -->avd.aquasec.com/nvd/cve-2020-8558  |
+-----------------------------+------------------+----------+--------------------+-------------------------+---------------------------------------+
```

```
$ trivy fs --severity HIGH ./go.sum
2022-01-10T17:33:01.529+0200    INFO    Number of language-specific files: 0
```

knqyf263 (commented):
Hi @jerbob92, thanks for the great PR! It looks awesome. I left a comment.

artifact/local/fs.go
jerbob92: Fixed, and added a test for both cases :)

jerbob92:
@knqyf263 should be good like this, don't know why I didn't add tests before.

knqyf263 (approved):

knqyf263:
Thanks for the great contribution!

knqyf263:
If you don't mind, could you submit a PR to Trivy as well?

jerbob92:
@knqyf263 Sure, will do tomorrow!

PR #357, opened 2022-01-07
Add pip compile support

jerbob92:
This PR extends the python/pip analyzer to allow for `pip-compile` file combinations.
`pip-compile` works by having an `.in` file, which would contain what you would normally have in your `requirements.txt` file, and is able to generate a `requirements.txt` like file from that but with exact versions and optionally things like hashes of the packages to prevent external tampering with the packages.

While fanal would generally already support it if the files would be named `requirements.in` and `requirements.txt`, however, it's common to split them up into multiple files, like `dev.in`/`dev.txt` and `prod.in`/`prod.txt`, this change is able to find pairs for `.in`/`.txt`. To support that I had to add the possibility to pass the directory into the analyzer `Required()` method, which causes this PR to be quite large. Let me know what you think.

Dependabot also looks for files like this.


analyzer/language/python/pip/pip.go
```diff
@@ -30,9 +32,24 @@ func (a pipLibraryAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisIn
...
+		return true
+	}
+
+	// Test for .txt/.in file combinations, which indicate pip-compile files.
+	if dir != "" && strings.HasSuffix(fileName, ".txt") {
+		// Create new filepath that has the same path but ends with .in
+		inFilePath := strings.TrimSuffix(filePath, ".txt") + ".in"
+
+		// Check whether in file exists.
+		if _, err := os.Stat(path.Join(dir, inFilePath)); err == nil {
```
knqyf263, line 47: Unfortunately, it doesn't work for container images.

jerbob92: @knqyf263 Is there a way to get this to work in docker though?

PR #367, opened 2022-01-11
fix(rpm): do not ignore installed files via third-party rpm

DmitriyLewen:
## Description
Added checking of rpm packages for being provided by OS vendor. 
## Issues 
- aquasecurity/trivy/issues/1546
- aquasecurity/trivy/issues/1481

### DIFF
--- a/analyzer/pkg/rpm/rpm.go
+++ b/analyzer/pkg/rpm/rpm.go
@@ -30,7 +30,18 @@ var requiredFiles = []string{
 }
 var errUnexpectedNameFormat = xerrors.New("unexpected name format")
 
-var osVendors = []string{"Amazon Linux", "Amazon.com", "CentOS", "Fedora Project", "Oracle America", "Red Hat, Inc.", "AlmaLinux", "CloudLinux"}
+var osVendors = []string{
+	"Amazon Linux",     // Amazon Linux 1
+	"Amazon.com",       // Amazon Linux 2
+	"CentOS",           // CentOS
+	"Fedora Project",   // Fedora
+	"Oracle America",   // Oracle Linux
+	"Red Hat, Inc.",    // Red Hat
+	"AlmaLinux",        // AlmaLinux
+	"CloudLinux",       // AlmaLinux
+	"VMware",           // Photon
+	"openSUSE Project", // OpenSUSE
+}
 
 type rpmPkgAnalyzer struct{}
 
@@ -108,7 +119,7 @@ func (a rpmPkgAnalyzer) parsePkgInfo(rc io.Reader) ([]types.Package, []string, e
 
 		//check if the package is preinstalled on the system
 		var files []string
-		if utils.StringInSlice(pkg.Vendor, defaultOsVendors) {
+		if utils.StringInSlice(pkg.Vendor, osVendors) {
 			files, err = pkg.InstalledFiles()
 			if err != nil {
 				return nil, nil, xerrors.Errorf("unable to get installed files: %w", err)
--- a/analyzer/pkg/rpm/rpm.go
+++ b/analyzer/pkg/rpm/rpm.go
@@ -191,7 +191,7 @@ func (a rpmPkgAnalyzer) Version() int {
 
 func packageProvidedByVendor(pkgVendor string) bool {
 	for _, vendor := range osVendors {
-		if strings.Contains(pkgVendor, vendor) {
+		if strings.HasPrefix(pkgVendor, vendor) {
 			return true
 		}
 	}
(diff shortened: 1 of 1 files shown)


analyzer/pkg/rpm/rpm.go
```diff
@@ -104,9 +106,13 @@ func (a rpmPkgAnalyzer) parsePkgInfo(rc io.Reader) ([]types.Package, []string, e
 			}
 		}
 
-		files, err := pkg.InstalledFiles()
-		if err != nil {
-			return nil, nil, xerrors.Errorf("unable to get installed files: %w", err)
+		//check if the package is preinstalled on the system
+		files := []string{}
```
knqyf263, line 110: nil slice is preferred.

```suggestion
		var files []string
```
https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

analyzer/pkg/rpm/rpm.go
```diff
@@ -30,6 +30,8 @@ var requiredFiles = []string{
 }
 var errUnexpectedNameFormat = xerrors.New("unexpected name format")
 
+var defaultOsVendors = []string{"Amazon Linux", "Amazon.com", "CentOS", "Fedora Project", "Oracle America", "Red Hat, Inc.", "AlmaLinux", "CloudLinux"}
```
knqyf263, line 33: "default" may be a bit confusing. `osVendors` looks enough to me.

```suggestion
var osVendors = []string{"Amazon Linux", "Amazon.com", "CentOS", "Fedora Project", "Oracle America", "Red Hat, Inc.", "AlmaLinux", "CloudLinux"}
```

knqyf263: Could you add which OS the vendor refers to as a comment so that we can easily identify which OS is missing? Looks like Photon OS and SUSE are missing.

```suggestion
var defaultOsVendors = []string{
    "Amazon Linux", // Amazon Linux 1
    "Amazon.com", // Amazon Linux 2
    "CentOS", // CentOS
    "Fedora Project", // Fedora
    "Oracle America", // Oracle Linux
    "Red Hat, Inc.", // Red Hat
    "AlmaLinux", // AlmaLinux
    "CloudLinux" // ??
}
```

knqyf263 (changes requested):

analyzer/pkg/rpm/rpm.go
```diff
@@ -104,9 +106,13 @@ func (a rpmPkgAnalyzer) parsePkgInfo(rc io.Reader) ([]types.Package, []string, e
 			}
 		}
 
-		files, err := pkg.InstalledFiles()
-		if err != nil {
-			return nil, nil, xerrors.Errorf("unable to get installed files: %w", err)
+		//check if the package is preinstalled on the system
```
knqyf263, line 109: I think this is not a pre-installed package. You may install a new package provided by the vendor later.

```suggestion
		// check if the package is vendor-provided
```

analyzer/pkg/rpm/rpm.go
```diff
@@ -30,6 +30,19 @@ var requiredFiles = []string{
...
+var osVendors = []string{
+	"Amazon Linux",     // Amazon Linux 1
+	"Amazon.com",       // Amazon Linux 2
+	"CentOS",           // CentOS
+	"Fedora Project",   // Fedora
+	"Oracle America",   // Oracle Linux
+	"Red Hat, Inc.",    // Red Hat
+	"AlmaLinux",        // AlmaLinux
+	"CloudLinux",       // AlmaLinux
+	"VMware",           // Photon
```
knqyf263, line 42: I found `VMware, Inc.` in `photon:3.0` In which version did you find it?

analyzer/pkg/rpm/rpm.go
```diff
@@ -30,6 +30,19 @@ var requiredFiles = []string{
...
+	"openSUSE Project", // OpenSUSE
```
knqyf263, line 43: I also found `SUSE LLC <https://www.suse.com/>` in `registry.suse.com/suse/sle15:15.3`, but there may be more variety of vendors in SLE. We may want to use `strings.HasPrefix` or `strings.Contains`.

analyzer/pkg/rpm/rpm.go
DmitriyLewen: That was my fault. `VMware, Inc.` is the right vendor. Thanks for finding this. I changed this.

analyzer/pkg/rpm/rpm.go
DmitriyLewen: Done

knqyf263 (changes requested):

knqyf263 (approved):

PR #369, opened 2022-01-11
fix(applier): Modify apply layer to merge custom resources as well

ankk13:
With addition for analyzer group, a CustomResource identifier is added to support any custom analyzer.
This commit allows merging of custom resources

### DIFF
--- a/applier/applier_test.go
+++ b/applier/applier_test.go
@@ -701,12 +701,10 @@ func TestApplier_ApplyLayers(t *testing.T) {
 			}
 
 			sort.Slice(got.CustomResources, func(i, j int) bool {
-				if got.CustomResources[i].FilePath < got.CustomResources[j].FilePath {
-					return true
-				} else if got.CustomResources[i].FilePath == got.CustomResources[j].FilePath {
+				if got.CustomResources[i].FilePath == got.CustomResources[j].FilePath {
 					return got.CustomResources[i].Type < got.CustomResources[j].Type
 				}
-				return false
+				return got.CustomResources[i].FilePath < got.CustomResources[j].FilePath
 			})
 
 			assert.Equal(t, tt.want, got)
--- a/applier/applier_test.go
+++ b/applier/applier_test.go
@@ -551,6 +551,13 @@ func TestApplier_ApplyLayers(t *testing.T) {
 										data: "Common Package type-B var/lib/dpkg/status.d/tzdata",
 									},
 								},
+								{
+									Type:     "type-C",
+									FilePath: "var/lib/dpkg/status.d/tzdata",
+									Data: dummyData{
+										data: "Common Package type-C, overidden in next layer",
+									},
+								},
 							},
 						},
 					},
@@ -607,6 +614,13 @@ func TestApplier_ApplyLayers(t *testing.T) {
 									FilePath: "php-app/composer.lock",
 									Data: dummyData{
 										data: "Common Application type-B php-app/composer.lock",
+									},
+								},
+								{
+									Type:     "type-C",
+									FilePath: "var/lib/dpkg/status.d/tzdata",
+									Data: dummyData{
+										data: "Type C application which replaces earlier detected resource",
 									},
 								},
 							},
@@ -689,6 +703,17 @@ func TestApplier_ApplyLayers(t *testing.T) {
 						},
 						Data: dummyData{
 							data: "Common Package type-B var/lib/dpkg/status.d/tzdata",
+						},
+					},
+					{
+						Type:     "type-C",
+						FilePath: "var/lib/dpkg/status.d/tzdata",
+						Layer: types.Layer{
+							Digest: "sha256:dffd9992ca398466a663c87c92cfea2a2db0ae0cf33fcb99da60eec52addbfc5",
+							DiffID: "sha256:aad63a9339440e7c3e1fff2b988991b9bfb81280042fa7f39a5e327023056819",
+						},
+						Data: dummyData{
+							data: "Type C application which replaces earlier detected resource",
 						},
 					},
 					{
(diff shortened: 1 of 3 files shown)


knqyf263 (changes requested):
Could you add a test case having CustomResource? Also, the file path should conflict with OS packages or language-specific packages so that we can confirm duplicate file paths work properly.

ankk13:
> Could you add a test case having CustomResource? Also, the file path should conflict with OS packages or language-specific packages so that we can confirm duplicate file paths work properly.

I have updated PR with the required unit tests.

ankk13:
> Could you add a test case having CustomResource? Also, the file path should conflict with OS packages or language-specific packages so that we can confirm duplicate file paths work properly.

I have updated PR with the required unit tests.

knqyf263 (changes requested):

applier/applier_test.go
```diff
@@ -500,6 +504,219 @@ func TestApplier_ApplyLayers(t *testing.T) {
...
+										data: "Common Application type-A php-app/composer.lock",
+									},
+								},
+								{
+									Type:     "type-B",
+									FilePath: "php-app/composer.lock",
+									Data: dummyData{
+										data: "Common Application type-B php-app/composer.lock",
+									},
+								},
```
knqyf263, line 611: We should check if replacing a previous layer works or not. When the type and file path are the same, the upper layer must replace CustomResources in the previous layer. In this case, the previous layer has `"var/lib/dpkg/status.d/tzdata"` with `type-B`. We should have the same pair with different data.

knqyf263 (changes requested):

applier/applier_test.go
```diff
@@ -500,6 +506,246 @@ func TestApplier_ApplyLayers(t *testing.T) {
...
+									Libraries: []types.Package{
+										{
+											Name:    "guzzlehttp/guzzle",
+											Version: "6.2.0",
+										},
+										{
+											Name:    "symfony/process",
+											Version: "v4.2.7",
+										},
+									},
```
knqyf263, line 603: Why do we need this object?

applier/applier_test.go
```diff
@@ -500,6 +506,246 @@ func TestApplier_ApplyLayers(t *testing.T) {
...
+									Data: dummyData{
+										data: "Common Package type-A var/lib/dpkg/status.d/tzdata",
+									},
+								},
+								{
+									Type:     "type-B",
+									FilePath: "var/lib/dpkg/status.d/tzdata",
+									Data: dummyData{
+										data: "Common Package type-B var/lib/dpkg/status.d/tzdata",
+									},
```
knqyf263, line 554: What is this for? We have `type-C` and it seems to be enough.

knqyf263 (changes requested):

applier/applier_test.go
ankk13: This case was added for case, if same resource on same layer is added by different analyzers, they should be equivalently merged. Tesecase check for intralayer merging. 
Type-C case is specific to same resource, on **different** layer by different analyzer.

knqyf263: >if same resource on same layer is added by different analyzers

Type-C seems to satisfy the above condition.

We want to test the following cases.

1. The same resource on the same layer by different analyzers
2. The same resource on the different layers by the same analyzer

The pair of Type-A and Type-C seems to cover both. Or, am I missing something? Can Type-B test something new?

ankk13: Above mentioned testcases are satisfied by Type-A and Type-C. 
But following testcase is missed. 
Consider a single layer with same resource(1 as Type-A and 1 as Type-C) but detected by different analyzer. If we have bug introduced in it merging this data. The above testcase may mask this failure, as the data will be overridden by merge on the next layer.

ankk13: Above mentioned testcases are satisfied by Type-A and Type-C. 
But following testcase is missed. 
Consider a single layer with same resource(1 as Type-A and 1 as Type-C) but detected by different analyzer. If we have bug introduced in it merging this data(intralayer). The above testcase may mask this failure, as the data will be overridden by merge on the next layer.

knqyf263: What if we move Type-A to the upper layer?

- upper layer
  - foo.txt
    - Type-A
    - Type-C
- lower layer
  - foo.txt
    - Type-C

ankk13: okey. I will update it accordingly

knqyf263 (changes requested):

applier/applier_test.go
```diff
@@ -500,6 +506,174 @@ func TestApplier_ApplyLayers(t *testing.T) {
...
+							CustomResources: []types.CustomResource{
+								{
+									Type:     "type-A",
+									FilePath: "var/lib/dpkg/status.d/tzdata",
+									Data: dummyData{
+										data: "Common Package type-A var/lib/dpkg/status.d/tzdata",
+									},
+								},
+								{
+									Type:     "type-C",
```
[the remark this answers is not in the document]
ankk13: Done

applier/applier_test.go
```diff
@@ -525,6 +699,16 @@ func TestApplier_ApplyLayers(t *testing.T) {
...
 				})
 			}
+
+			sort.Slice(got.CustomResources, func(i, j int) bool {
+				if got.CustomResources[i].FilePath < got.CustomResources[j].FilePath {
+					return true
+				} else if got.CustomResources[i].FilePath == got.CustomResources[j].FilePath {
+					return got.CustomResources[i].Type < got.CustomResources[j].Type
+				}
+				return false
```
knqyf263, line 709: How about this?
```suggestion
				if  got.CustomResources[i].FilePath == got.CustomResources[j].FilePath {
					return got.CustomResources[i].Type < got.CustomResources[j].Type
				}
				return got.CustomResources[i].FilePath < got.CustomResources[j].FilePath
```

ankk13: yes, more readable. Updated

knqyf263 (approved):
:shipit:

PR #370, opened 2022-01-12
Extend Dockerfile filename support

jerbob92:
This PR changes the Dockerfile matching to a regex, documents the allowed fileformats and adds support for filenames like Dockerfile-build.

### DIFF
--- a/analyzer/config/docker/docker.go
+++ b/analyzer/config/docker/docker.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strings"
 
 	"golang.org/x/xerrors"
 
@@ -16,7 +15,14 @@ import (
 
 const version = 1
 
-var requiredFile = "Dockerfile"
+/*
+  This regex allows for the following filename formats:
+  - Dockerfile (default)
+  - Dockerfile.build (with Dockerfile as filename and environment as extension)
+  - build.dockerfile (with environment as filename and dockerfile as extension)
+  - Dockerfile-build (environment appended with dash to Dockerfile)
+*/
+var fileRegex = regexp.MustCompile(`(?i)^(.*[\.])?Dockerfile([-\.].*)?$`)
 
 type ConfigAnalyzer struct {
 	parser      *dockerfile.Parser
@@ -54,16 +60,8 @@ func (s ConfigAnalyzer) Required(filePath string, _ os.FileInfo) bool {
 		return true
 	}
 
-	base := filepath.Base(filePath)
-	ext := filepath.Ext(base)
-	if strings.EqualFold(base, requiredFile+ext) {
-		return true
-	}
-	if strings.EqualFold(ext, "."+requiredFile) {
-		return true
-	}
-
-	return false
+	fileName := filepath.Base(filePath)
+	return fileRegex.MatchString(fileName)
 }
 
 func (s ConfigAnalyzer) Type() analyzer.Type {
--- a/analyzer/config/docker/docker_test.go
+++ b/analyzer/config/docker/docker_test.go
@@ -208,6 +208,11 @@ func Test_dockerConfigAnalyzer_Required(t *testing.T) {
 			want:     true,
 		},
 		{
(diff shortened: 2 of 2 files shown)


jerbob92:
@knqyf263 Why do these benchmarks keep bothering me? :( I can't imagine a regex being much worse than what it was before.

knqyf263:
Thanks for your contribution! But we already have support for regexp. User can specify it.
https://github.com/aquasecurity/fanal/blob/f159d650c45ee54b6f1bf346200897dd0b1251c8/analyzer/config/docker/docker.go#L53-L55

knqyf263:
>@knqyf263 Why do these benchmarks keep bothering me? :( I can't imagine a regex being much worse than what it was before.
 
Never mind. It used to be stable, but recently unstable. I guess it depends on the environment of GitHub Actions.

jerbob92:
> Thanks for your contribution! But we already have support for regexp. User can specify it.
> 
> https://github.com/aquasecurity/fanal/blob/f159d650c45ee54b6f1bf346200897dd0b1251c8/analyzer/config/docker/docker.go#L53-L55

I know that people can specify their own regex. The reason I'm adding all these extra filename formats is to include as much files in the scan that should be scanned by default.

knqyf263:
>The reason I'm adding all these extra filename formats is to include as much files in the scan that should be scanned by default.

Yes, I assumed so, but it also increases false detection as well. They may not want to scan `Dockerfile.test` or something like that. I prefer strict by default and flexible by user-specified.

jerbob92:
> > The reason I'm adding all these extra filename formats is to include as much files in the scan that should be scanned by default.
> 
> Yes, I assumed so, but it also increases false detection as well. They may not want to scan `Dockerfile.test` or something like that. I prefer strict by default and flexible by user-specified.

If they do not want that they can exclude the files. It honestly makes no sense to implement a custom file regex per scan type.

knqyf263:
I'm sorry, but it honestly makes no sense to have false detection and exclude them. Our policy is that security scanner should not be noisy.

jerbob92:
> I'm sorry, but it honestly makes no sense to have false detection and exclude them. Our policy is that security scanner should not be noisy.

You know that Dockerfile.test and test.dockerfile are both currently accepted filenames right?

knqyf263:
Yeah, maybe it is not a good example.

knqyf263:
Conversely, the examples you listed are already covered.

jerbob92:
> Conversely, the examples you listed are already covered.

Yes, I improved the code by adding documentation which formats are supported and converted the existing code into a regex to match all current accepted formats + the one that I wanted (Dockerfile-{suffix})

jerbob92:
But I see your point. However, the existing logic then isn't too great, because not every scanner supports this regex.
Would you accept a PR that implements the regex matching on a higher level before the `analyzer.Required()` method is called? So that we can support it automatically on every analyzer?

knqyf263:
If I understand correctly, your regexp matches `foo.Dockerfile-bar`, right?

jerbob92:
> If I understand correctly, your regexp matches `foo.Dockerfile-bar`, right?

That is correct.

knqyf263:
>However, the existing logic then isn't too great, because not every scanner supports this regex.

The config analyzers are a bit special since they can be any file name. All the config analyzers support the regex. 

From the perspective of vulnerability scanning, the required file names are basically fixed. No need for regexp.

jerbob92:
> > However, the existing logic then isn't too great, because not every scanner supports this regex.
> 
> The config analyzers are a bit special since they can be any file name. All the config analyzers support the regex.
> 
> From the perspective of vulnerability scanning, the required file names are basically fixed. No need for regexp.

That's not really true, see my PR's #355 and #357.

knqyf263:
Hmm. Interesting. Need to consider that.

knqyf263:
Let's see how many users want this kind of capability. You can open an issue in Trivy and we can see the reaction of the community.

Actually, I agree that specifying regexp per file type is complex. If we support regexp for vulnerability scanning, we have to think about UI/UX.

PR #372, opened 2022-01-12
Move file patterns to a higher level to be able to use it on any analyzer

jerbob92:
This PR moves the file patterns feature from the config analyzers to a global level so that we can use it on any analyzer.
In favor of #370, #357, #355 as this fill allow you to set the needed file patterns yourself. 

This also makes the config analyzers more like the other analyzers.

We talked about asking the community, but IMHO it makes sense to just add it as it is a logical next step after having it in the config analyzers.

This can be enabled in Trivy without changing anything to the config scanner. The other commands will just get a `--filte-patterns` option, I already have a branch for that ready.

What do you think @knqyf263?

### DIFF
--- a/analyzer/all/import.go
+++ b/analyzer/all/import.go
@@ -2,6 +2,13 @@ package all
 
 import (
 	_ "github.com/aquasecurity/fanal/analyzer/command/apk"
+	_ "github.com/aquasecurity/fanal/analyzer/config/cloudformation"
+	_ "github.com/aquasecurity/fanal/analyzer/config/docker"
+	_ "github.com/aquasecurity/fanal/analyzer/config/hcl"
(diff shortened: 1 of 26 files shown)


jerbob92:
@knqyf263 could you please take a look?

jerbob92:
@knqyf263 could you please take a look?

jerbob92:
@knqyf263 could you please take a look?

jerbob92:
@knqyf263 could you please take a look?

jerbob92:
@knqyf263 I have updated the MR to the latest version of Fanal

jerbob92:
@knqyf263 for how long do I have to maintain this MR until it gets merged?

jerbob92:
image

@knqyf263

PR #406, opened 2022-02-21
chore: update the defsec and tfsec versions

owenrumney:
- rework some of the iac tests to be more flexible to change
- update the scanner to use the moved parser
  - everything is now in defsec now for CF parsing, scanning and testing


liamg (approved):
LGTM :rocket:

owenrumney:
> Looks good, but I'm not familiar with defsec now. I'd like to wait for @liamg's review.

short version is that defsec holds all the rules and the parsing capability for Terraform and CloudFormation. It also has the scanner for CloudFormation now that there is no appetite for a dedicated CF tool

PR #419, opened 2022-03-03
fix(filter-hook): systemfiles filter for custom resources

ankk13:
Currently Filter hook filters systemfiles resources from Applications and Packages.
Updated the hook to filter CustomResources as well.


knqyf263:
Please try to keep the subject line to 72 characters or less and write the detail in the description.

knqyf263 (approved):

knqyf263:
:shipit:

PR #425, opened 2022-03-17
fix(docker): getting an image by ID or a name with tag

afdesk:
`fanal` tries to find an image by name (with tag). 
If it's not found, `fanal` will look for an image by ID.

Fixes a problem from #405

### DIFF
--- a/image/daemon/docker.go
+++ b/image/daemon/docker.go
@@ -24,7 +24,10 @@ func DockerImage(ref name.Reference) (Image, func(), error) {
 		}
 	}()
 
-	imageID := ref.Name() // <image_name>:<tag> pattern like "alpine:3.15"
+	// <image_name>:<tag> pattern like "alpine:3.15"
+	// or
+	// <image_name>@<digest> pattern like "alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300"
+	imageID := ref.Name()
 	inspect, _, err := c.ImageInspectWithRaw(context.Background(), imageID)
 	if err != nil {
 		imageID = ref.String() // <image_id> pattern like `5ac716b05a9c`


image/daemon/docker.go
```diff
@@ -24,10 +24,14 @@ func DockerImage(ref name.Reference) (Image, func(), error) {
...
 
-	imageID := ref.String()
+	imageID := ref.Name() // Image ID uses "<Image_name>:<Tag>" pattern
 	inspect, _, err := c.ImageInspectWithRaw(context.Background(), imageID)
 	if err != nil {
-		return nil, cleanup, xerrors.Errorf("unable to inspect the image (%s): %w", imageID, err)
+		imageID = ref.String() // Image ID uses "<Image_ID>" pattern
+		inspect, _, err = c.ImageInspectWithRaw(context.Background(), imageID)
+		if err != nil {
+			return nil, cleanup, xerrors.Errorf("unable to inspect the image (%s): %w", ref.Name(), err)
```
knqyf263, line 33: `ref.String()` returns an original value. I think `ref.String()` is better here.

```suggestion
			return nil, cleanup, xerrors.Errorf("unable to inspect the image (%s): %w", imageID, err)
```

image/daemon/docker.go
```diff
@@ -24,10 +24,14 @@ func DockerImage(ref name.Reference) (Image, func(), error) {
 		}
 	}()
 
-	imageID := ref.String()
+	imageID := ref.Name() // Image ID uses "<Image_name>:<Tag>" pattern
```
knqyf263, line 27:
```suggestion
	// <image_name>:<tag> pattern like "alpine:3.15"
	// or
	// <image_name>@<digest> pattern like "alpine@sha256:21a3deaa0d32a8057914f36584b5288d2e5ecc984380bc0118285c70fa8c9300"
	imageID := ref.Name() 
```

image/daemon/docker.go
```diff
@@ -24,10 +24,14 @@ func DockerImage(ref name.Reference) (Image, func(), error) {
...
 	inspect, _, err := c.ImageInspectWithRaw(context.Background(), imageID)
 	if err != nil {
-		return nil, cleanup, xerrors.Errorf("unable to inspect the image (%s): %w", imageID, err)
+		imageID = ref.String() // Image ID uses "<Image_ID>" pattern
```
knqyf263, line 30: nit

```suggestion
		imageID = ref.String() // <image_id> pattern like `5ac716b05a9c`
```

knqyf263 (changes requested):
Please refer to the original PR.

image/daemon/docker.go
afdesk: done

image/daemon/docker.go
afdesk: don

image/daemon/docker.go
afdesk: done

image/daemon/docker.go
knqyf263: You missed my suggested change.

knqyf263: I mean digest

afdesk: sorry, done

PR #426, opened 2022-03-21
feat(cache): add DeleteBlobs to ArtifactCache

afdesk:


cache/s3.go
```diff
@@ -39,6 +39,14 @@ func (c S3Cache) PutArtifact(artifactID string, artifactConfig types.ArtifactInf
 	return nil
 }
 
+func (c S3Cache) DeleteBlob(blobID string) error {
+	err := c.DeleteBlob(blobID)
```
knqyf263, line 43: Looks like infinite loop. Am I missing something?

afdesk: sure, my mistake. fixed

artifact/image/image.go
```diff
@@ -120,6 +120,10 @@ func (a Artifact) Inspect(ctx context.Context) (types.ArtifactReference, error)
 	}, nil
 }
 
+func (_ Artifact) Clean(_ types.ArtifactReference) error {
```
knqyf263, line 123: nit

```suggestion
func (Artifact) Clean(_ types.ArtifactReference) error {
```

artifact/remote/git.go
```diff
@@ -81,6 +81,10 @@ func (a Artifact) Inspect(ctx context.Context) (types.ArtifactReference, error)
 	return ref, nil
 }
 
+func (_ Artifact) Clean(_ types.ArtifactReference) error {
```
knqyf263, line 84:
```suggestion
func (Artifact) Clean(_ types.ArtifactReference) error {
```

cache/cache.go
```diff
@@ -28,6 +28,9 @@ type ArtifactCache interface {
 
 	// PutBlob stores blob information such as layer information in local cache
 	PutBlob(blobID string, blobInfo types.BlobInfo) (err error)
+
+	// DeleteBlob removes blob by Id
+	DeleteBlob(blobID string) error
```
knqyf263, line 33: After some consideration, I feel like it should take a slice.

```suggestion
	DeleteBlobs(blobIDs []string) error
```

cache/redis.go
```diff
@@ -50,6 +50,14 @@ func (c RedisCache) PutBlob(blobID string, blobInfo types.BlobInfo) error {
 	}
 	return nil
 }
+func (c RedisCache) DeleteBlob(blobID string) error {
```
knqyf263, line 53: Want to add tests

artifact/image/image.go
afdesk: done

artifact/remote/git.go
afdesk: done

cache/cache.go
afdesk: done

cache/fs.go
```diff
@@ -113,6 +115,21 @@ func (fs FSCache) GetArtifact(artifactID string) (types.ArtifactInfo, error) {
...
+	fs.db.Update(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket([]byte(blobBucket))
+		for _, blobID := range blobIDs {
+			if err := bucket.Delete([]byte(blobID)); err != nil {
+				errs = multierror.Append(errs, err)
+			}
+		}
+		return nil
+	})
+	return errs
```
knqyf263, line 130: Also, we should handle an error from `Update`.

```suggestion
	err := fs.db.Update(func(tx *bolt.Tx) error {
		bucket := tx.Bucket([]byte(blobBucket))
		for _, blobID := range blobIDs {
			if err := bucket.Delete([]byte(blobID)); err != nil {
				errs = multierror.Append(errs, err)
			}
		}
		return nil
	})
	if err != nil {
		return xerrors.Errorf("DB update error: %w", err)
	}
	return errs
```

knqyf263 (changes requested):

cache/fs.go
afdesk: done

afdesk: oh @knqyf263 I missed.
is it really possible that `fs.db.Update` will return any error in this case?

afdesk: we accumulate all inner errors into `errs`...

knqyf263: Yes, it is possible to return an error. For example, the database is locked and db.Update can't open the database for write operations. The database is broken for some reason. etc...

cache/redis.go
afdesk: added

cache/s3.go
```diff
@@ -39,6 +39,16 @@ func (c S3Cache) PutArtifact(artifactID string, artifactConfig types.ArtifactInf
 	return nil
 }
 
+func (c S3Cache) DeleteBlob(blobID string) error {
```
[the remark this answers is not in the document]
afdesk: added tests

cache/fs.go
afdesk: oh, sure. thanks for explanation!

cache/redis_test.go
```diff
@@ -500,3 +502,62 @@ func TestRedisCache_Clear(t *testing.T) {
...
+				addr = "dummy:16379"
+			}
+
+			c := cache.NewRedisCache(&redis.Options{
+				Addr: addr,
+			})
+
+			err = c.DeleteBlobs(tt.args.blobIDs)
+			if tt.wantErr != "" {
+				require.NotNil(t, err)
```
knqyf263, line 555: nit
```suggestion
				require.Error(t, err)
```

cache/redis_test.go
```diff
@@ -500,3 +502,62 @@ func TestRedisCache_Clear(t *testing.T) {
...
+				assert.Contains(t, err.Error(), tt.wantErr)
+				return
+			} else {
+				assert.NoError(t, err)
+			}
```
knqyf263, line 560: nit
```suggestion
			}
			assert.NoError(t, err)
```

artifact/local/fs.go
```diff
@@ -165,3 +165,7 @@ func (a Artifact) Inspect(ctx context.Context) (types.ArtifactReference, error)
 		BlobIDs: []string{cacheKey},
 	}, nil
 }
+
+func (a Artifact) Clean(reference types.ArtifactReference) error {
+	return a.cache.DeleteBlobs([]string{reference.ID})
```
knqyf263, line 170: It should be reference.BlobIDs technically. In filesystem scanning, `artifactID` and `blobIDs` are the same "by chance". They are usually different and the method called here is "DeleteBlobs". Passing `blobIDs` makes sense.
https://github.com/aquasecurity/fanal/blob/3519fe6e6c218e32a87b1912132d59891c171cfb/types/artifact.go#L94

afdesk: updated

PR #429, opened 2022-03-28
fix: handle no filename comeing through from defsec

owenrumney:
- fix: fix panic with single file location
- fix: catch results where the filename is not specified


knqyf263:
Oh, I see. I'm sorry for the lack of explanation. GitHub has "Squash and merge" button.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request

knqyf263:
image

PR #430, opened 2022-03-29
feat(ubuntu): added support for Ubuntu ESM versions

DmitriyLewen:
## Description
Ubuntu has [ESM program](https://ubuntu.com/security/esm).
ESM status information is stored in the`var/lib/ubuntu-advantage/status.json` file.
- added parsing `status.json` file.
- added function for merging OS versions.

## Blockers
- [ ] aquasecurity/trivy-db/pull/200


analyzer/analyzer.go
```diff
@@ -291,3 +286,32 @@ func isDisabled(t Type, disabled []Type) bool {
...
+			return old
+		}
+		if old.Name == esmVersionSuffix {
+			new.Name = new.Name + "-" + old.Name
+			return new
+		}
+		if new.Name == esmVersionSuffix {
+			new.Name = old.Name + "-" + new.Name
+			return new
+		}
```
knqyf263, line 314: It is a bit tricky. What if we add `Extended` field to `OS`? Then, we can use the field for Extended LTS (Debian), Extended Life Cycle Support (Red Hat), and so on.

DmitriyLewen: Done

DmitriyLewen: Done

PR #431, opened 2022-03-31
feat: add secret scanning

knqyf263:
## Description
It detects secrets such as AWS Access Key and GitHub PAT.

## Issues
- https://github.com/aquasecurity/trivy/issues/1950

## TODO
- [x] Add secret analyzer
- [x] Add secret scanner
  - [x] Support secret detection
  - [x] Support line numbers
  - [x] Mask secrets
  - [x] Support custom rules from a config file
- [x] Add built-in rules (https://github.com/aquasecurity/fanal/pull/464)
- [x] Support allow patterns (#443)
- [x] Support keywords (#472)
- [x] Optimization
    - [x] Skip some files, dirs and extensions
    - [x] Skip binaries
    - [x] Consider container image scanning
- [x] Add unit tests
  - [x] secret/scanner.go
  - [x] analyzer/secret/secret.go
- [x] Add integration tests (will be added in Trivy)

### DIFF
--- a/secret/builtin.go
+++ b/secret/builtin.go
@@ -743,6 +743,11 @@ var builtinRules = []Rule{
 }
 var builtinAllowRules = []AllowRule{
 	{
+		ID:          "tests",
+		Description: "Avoid paths containing test",
+		Path:        MustCompile(`\/test`),
+	},
+	{
 		ID:          "vendor",
 		Description: "Vendor dirs",
 		Path:        MustCompile(`\/vendor\/`),


secret/scanner_test.go
```diff
@@ -0,0 +1,116 @@
...
+			configPath:    "testdata/exclude-block.yaml",
+			inputFilePath: "testdata/secret.txt",
+			want: types.Secret{
+				FilePath: "testdata/secret.txt",
+				Findings: []types.SecretFinding{wantFinding2},
+			},
+		},
+		{
+			name:          "global allow-rule path",
+			configPath:    "testdata/global-allow-path.yaml",
```
VaismanLior, line 74: Do you think we should add another test to check that the global allow path works? 
Right now it checks that paths that shouldn't be allowed, are not allowed. But it doesn't cover the case if the feature is broken.

knqyf263: Do you mean it doesn't cover the case when the scanning is completely broken and doesn't detect anything?

VaismanLior: I mean that it doesn't check that the `global allow-rule path` skips any secrets.
The path `foo` doesn't exist so all secrets are detected, but at the same time, the `global allow-rule path` could not work and it will detect all the secrets as well.

VaismanLior: Cool, I'll fix that.

VaismanLior (approved):

PR #443, opened 2022-04-03
Added secrets allow list

VaismanLior:
Added allowlist options for every rule and a global allowlist(all rules).
Allowlist can contain:
 - Regexes 
 - Paths
Any secret that is found inside the allowed pattern will be ignored
Global allow list is from [gitleaks default config](https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml)

### DIFF
--- a/secret/builtin.go
+++ b/secret/builtin.go
@@ -34,31 +34,18 @@ var builtinRules = []Rule{
 		Severity: "CRITICAL",
 		Regex:    regexp.MustCompile(`gho_[0-9a-zA-Z]{36}`),
 	},
-	{
-		ID:       "generic-secret",
-		Type:     RuleTypeGenericSecret,
-		Title:    "Generic Secret",
-		Severity: "HIGH",
-		Regex:    regexp.MustCompile(`(?i)(?:dbpasswd|dbhost|token|secret|password|pwd)\w*["']?\]?\s*[=:]\s*['"]([\w@0-9a-zA-Z-_\/+!{}\/=. ]{5,160})['"]`),
-		AllowList: AllowList{
-			Title: "False positives",
-			Paths: []*regexp.Regexp{regexp.MustCompile(`(.*?)(\.env|gradle\.properties)$`)},
-			Regexes: []*regexp.Regexp{
-				regexp.MustCompile(`(?i)\w*(?:prefix|suffix|default|field|key)\w*\s*[=:]`),
-				regexp.MustCompile(`(?i)\w*(prefix|suffix|field|default|key|$\{.*?\})\w*[\]"']*\s+[=:]`),
-				regexp.MustCompile(`(?i)=[\s*\w\S]*[\[\(][[\s\S\w]*[\]\)]`),
-				regexp.MustCompile(`(?i)=\s*(\w*[\.\/]\w*)+`),
-				regexp.MustCompile(`(?i)[=:]\s*['"](\w*[\.\/]\w*)+['"]`),
-				regexp.MustCompile(`(?i)[=:]\s*\S*(sample|example|false|true|some)`),
-				regexp.MustCompile(`(?i)ENC[[]`),
-				regexp.MustCompile(`xox[baprs]-([0-9a-zA-Z]{10,48})`),
-				regexp.MustCompile(`(?i)(.{0,20})?['"][0-9a-f]{32}-us[0-9]{1,2}['"]`),
-				regexp.MustCompile(`(A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}`),
-				regexp.MustCompile(`AIza[0-9A-Za-z\\-_]{35}`),
-				regexp.MustCompile(`[:=]\s*\w*?(\.\w*)+`),
-				regexp.MustCompile(`[:=]\s*\w*?([\[\(]['"]?\w*['"]?[\]\)])+`),
-				regexp.MustCompile(`(?i)^\s*\[.*\]\s*[=:]`),
-			},
-		},
+}
+
+var builtinGlobalAllowList = AllowList{
+	Title: "global allow list",
+	Regexes: []*regexp.Regexp{
+		regexp.MustCompile(`219-09-9999`),
+		regexp.MustCompile(`078-05-1120`),
+		regexp.MustCompile(`(9[0-9]{2}|666)-\d{2}-\d{4}`),
+	},
+	Paths: []*regexp.Regexp{
+		regexp.MustCompile(`gitleaks.toml`),
+		regexp.MustCompile(`(.*?)(jpg|gif|doc|pdf|bin|svg|socket)$`),
+		regexp.MustCompile(`(go.mod|go.sum)$`),
 	},
 }
--- a/secret/builtin.go
+++ b/secret/builtin.go
@@ -7,8 +7,9 @@ import (
 )
 
 var (
-	RuleTypeAWS    = types.SecretRuleType("AWS")
-	RuleTypeGitHub = types.SecretRuleType("GitHub")
+	RuleTypeAWS           = types.SecretRuleType("AWS")
+	RuleTypeGitHub        = types.SecretRuleType("GitHub")
+	RuleTypeGenericSecret = types.SecretRuleType("GenericSecret")
 )
 
 var builtinRules = []Rule{
@@ -32,5 +33,32 @@ var builtinRules = []Rule{
 		Title:    "GitHub OAuth Access Token",
 		Severity: "CRITICAL",
 		Regex:    regexp.MustCompile(`gho_[0-9a-zA-Z]{36}`),
+	},
+	{
+		ID:       "generic-secret",
+		Type:     RuleTypeGenericSecret,
+		Title:    "Generic Secret",
+		Severity: "HIGH",
+		Regex:    regexp.MustCompile(`(?i)(?:dbpasswd|dbhost|token|secret|password|pwd)\w*["']?\]?\s*[=:]\s*['"]([\w@0-9a-zA-Z-_\/+!{}\/=. ]{5,160})['"]`),
+		AllowList: AllowList{
+			Title: "False positives",
+			Paths: []*regexp.Regexp{regexp.MustCompile(`(.*?)(\.env|gradle\.properties)$`)},
+			Regexes: []*regexp.Regexp{
+				regexp.MustCompile(`(?i)\w*(?:prefix|suffix|default|field|key)\w*\s*[=:]`),
+				regexp.MustCompile(`(?i)\w*(prefix|suffix|field|default|key|$\{.*?\})\w*[\]"']*\s+[=:]`),
+				regexp.MustCompile(`(?i)=[\s*\w\S]*[\[\(][[\s\S\w]*[\]\)]`),
+				regexp.MustCompile(`(?i)=\s*(\w*[\.\/]\w*)+`),
+				regexp.MustCompile(`(?i)[=:]\s*['"](\w*[\.\/]\w*)+['"]`),
+				regexp.MustCompile(`(?i)[=:]\s*\S*(sample|example|false|true|some)`),
+				regexp.MustCompile(`(?i)ENC[[]`),
+				regexp.MustCompile(`xox[baprs]-([0-9a-zA-Z]{10,48})`),
+				regexp.MustCompile(`(?i)(.{0,20})?['"][0-9a-f]{32}-us[0-9]{1,2}['"]`),
+				regexp.MustCompile(`(A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}`),
+				regexp.MustCompile(`AIza[0-9A-Za-z\\-_]{35}`),
+				regexp.MustCompile(`[:=]\s*\w*?(\.\w*)+`),
+				regexp.MustCompile(`[:=]\s*\w*?([\[\(]['"]?\w*['"]?[\]\)])+`),
+				regexp.MustCompile(`(?i)^\s*\[.*\]\s*[=:]`),
+			},
+		},
 	},
 }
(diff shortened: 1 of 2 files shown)


secret/scanner.go
```diff
@@ -39,18 +48,75 @@ type ScanArgs struct {
...
 		}
 
+		// Find allowed locations
+		allowedLocations := make([][]int, 0)
+		if rule.AllowList.Regexes != nil && len(rule.AllowList.Regexes) > 0 {
+			if s.AllowList.Regexes != nil && len(s.AllowList.Regexes) > 0 {
+				rule.AllowList.Regexes = append(rule.AllowList.Regexes, s.AllowList.Regexes...)
+			}
+			for _, regex := range rule.AllowList.Regexes {
+				allowedLocations = append(allowedLocations, regex.FindAllIndex(args.Content, -1)...)
```
knqyf263, line 90: I don't want to apply allow rules to the entire file content. What if we apply allow rules to detected values only? In other words, the cost should be much smaller to apply them to `match`.

secret/builtin.go
```diff
@@ -34,3 +35,17 @@ var builtinRules = []Rule{
 		Regex:    regexp.MustCompile(`gho_[0-9a-zA-Z]{36}`),
 	},
 }
+
+var builtinGlobalAllowList = AllowList{
+	Title: "global allow list",
+	Regexes: []*regexp.Regexp{
+		regexp.MustCompile(`219-09-9999`),
+		regexp.MustCompile(`078-05-1120`),
```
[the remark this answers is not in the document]
VaismanLior: I looked it up, all three of them refer to American Social Security Numbers, they are common SSN examples.
We can make two of them use string comparison, or combine them all into one regex. 
Third option is just to remove them because I don't think they are common enough.
What do you think?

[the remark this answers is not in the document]
knqyf263: We can remove them first. If we see many complaints about them, we can add these values then.

secret/scanner.go
VaismanLior: Scanning the `match` itself will be faster but less effective.
When you scan for false positives in generic rules(not a specific secret form), many False Positives are detected by the surrounding code... I think this approach is more flexible.
It will also support cases of multiline regexs' in the allow list.
For example, if the detected "secret" is inside an `enum` declaration, we will be able to ignore it.

secret/builtin.go
```diff
@@ -34,3 +35,17 @@ var builtinRules = []Rule{
...
+		regexp.MustCompile(`(9[0-9]{2}|666)-\d{2}-\d{4}`),
+	},
+	Paths: []*regexp.Regexp{
+		regexp.MustCompile(`gitleaks.toml`),
+		regexp.MustCompile(`(.*?)(jpg|gif|doc|pdf|bin|svg|socket)$`),
+		regexp.MustCompile(`(go.mod|go.sum)$`),
+	},
```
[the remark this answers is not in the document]
VaismanLior: removed

secret/builtin.go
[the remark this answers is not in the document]
VaismanLior: 👍🏼

secret/scanner.go
```diff
@@ -39,18 +48,75 @@ type ScanArgs struct {
 
 func (s Scanner) Scan(args ScanArgs) types.Secret {
 	var findings []types.SecretFinding
+
+	if s.AllowList.Paths != nil && len(s.AllowList.Paths) > 0 {
```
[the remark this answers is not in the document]
VaismanLior: done

secret/scanner.go
```diff
@@ -39,18 +48,75 @@ type ScanArgs struct {
...
+		for _, path := range s.AllowList.Paths {
+			if path.MatchString(args.FilePath) {
+				return types.Secret{
+					FilePath: args.FilePath,
+					Findings: findings,
+				}
```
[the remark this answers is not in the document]
VaismanLior: done

secret/scanner.go
```diff
@@ -39,18 +48,75 @@ type ScanArgs struct {
...
 	for _, rule := range s.Rules {
 		// Check if the file path should be scanned by this rule
 		if rule.Path != nil && !rule.Path.MatchString(args.FilePath) {
 			continue
 		}
 
+		if rule.AllowList.Paths != nil && len(rule.AllowList.Paths) > 0 {
+			for _, path := range rule.AllowList.Paths {
+				if path.MatchString(args.FilePath) {
+					continue
```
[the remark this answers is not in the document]
VaismanLior: Missed that, fixed

secret/scanner.go
knqyf263: I'm confused since even generic rules capture values in `match` anyway.

>For example, if the detected "secret" is inside an enum declaration, we will be able to ignore it.

Would you elaborate on it?

knqyf263: Oh, my bad. I wanted to say the global allow rules should be evaluated once. Global allow rules are added here and it will be evaluated many times if multiple secrets are found.
https://github.com/aquasecurity/fanal/blob/622a930f2e4307a58235a1d57f1250582deed10d/secret/scanner.go#L82-L84

secret/scanner.go
```diff
@@ -39,31 +55,84 @@ type ScanArgs struct {
...
+		skipRule := false
+		for _, path := range rule.AllowList.Paths {
+			if path.MatchString(args.FilePath) {
+				skipRule = true
+				break
+			}
+		}
+		if skipRule {
+			continue
+		}
```
knqyf263, line 82: I think we can define like

```
func shouldSkipPath(filePath string, allowPaths []*regexp.Regexp) bool {
	for _, r := range allowPaths {
		if r.MatchString(filePath) {
			return true
		}
	}
	return false
}
```

 and

```suggestion
		if shouldSkipPath(args.FilePath, rule.AllowList.Paths) {
			continue
		}
```

secret/scanner.go
```diff
@@ -39,31 +55,84 @@ type ScanArgs struct {
...
+				Type:          rule.Type,
+				Severity:      strings.ToUpper(rule.Severity),
+				Title:         rule.Title,
+				StartPosition: loc[0],
+				EndPosition:   loc[1],
 			})
 		}
+
+		// Find rule allowed locations
+		allowedLocations := make([][]int, 0)
```
knqyf263, line 103: nit

```suggestion
		var allowedLocations [][]int
```

secret/scanner.go
```diff
@@ -72,6 +141,34 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
 	}
 }
 
+func removeAllowedSecrets(allRuleResults []RuleResult, allowedLocations [][]int) []RuleResult {
+	if len(allowedLocations) > 0 {
+		results := make([]RuleResult, 0)
+		for _, result := range results {
```
knqyf263, line 147: You may want to iterate allRuleResults here. Iterating `results` doesn't make sense.

secret/scanner.go
```diff
@@ -72,6 +141,34 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
...
+				}
+			}
+			if !found {
+				results = append(results, result)
+			}
+		}
+		return results
+	} else {
+		return allRuleResults
+	}
```
knqyf263, line 169: if-statement is large. The early return may be better.

```suggestion
	if len(allowedLocations) == 0 {
		return allRuleResults
	}
	
	results := make([]RuleResult, 0)
	for _, result := range results {
		found := false
		if result.EndPosition < allowedLocations[0][0] {
			results = append(results, result)
			continue
		}
		for _, allowedLocation := range allowedLocations {
		if allowedLocation[0] > result.EndPosition {
				break
			}
			if result.StartPosition >= allowedLocation[0] && result.EndPosition <= allowedLocation[1] {
				found = true
				break
			}
		}
		if !found {
			results = append(results, result)
		}
	}
	return results
```

knqyf263, line 169: if-statement is large. The early return may be better.

```suggestion
	if len(allowedLocations) == 0 {
		return allRuleResults
	}
	
	results := make([]RuleResult, 0)
	for _, result := range results {
		found := false
		if result.EndPosition < allowedLocations[0][0] {
			results = append(results, result)
			continue
		}
		for _, allowedLocation := range allowedLocations {
		if allowedLocation[0] > result.EndPosition {
				break
			}
			if result.StartPosition >= allowedLocation[0] && result.EndPosition <= allowedLocation[1] {
				found = true
				break
			}
		}
		if !found {
			results = append(results, result)
		}
	}
	return results
```

secret/scanner.go
```diff
@@ -72,6 +141,34 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
...
 
+func removeAllowedSecrets(allRuleResults []RuleResult, allowedLocations [][]int) []RuleResult {
+	if len(allowedLocations) > 0 {
+		results := make([]RuleResult, 0)
+		for _, result := range results {
+			found := false
+			if result.EndPosition < allowedLocations[0][0] {
+				results = append(results, result)
+				continue
+			}
```
knqyf263, line 152: Looks like we don't need these lines since we do the same thing at 154-156 lines.

secret/scanner.go
```diff
@@ -72,6 +141,34 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
...
+			for _, allowedLocation := range allowedLocations {
+				if allowedLocation[0] > result.EndPosition {
+					break
+				}
```
knqyf263, line 156: I think it assumes `allowedLocations` is sorted, but it looks like it is not the case. There are two places to call `removeAllowedSecrets()` and iterate `AllowList.Regexes`. The result of merging results of `regex.FindAllIndex()` will not be sorted even though `regex.FindAllIndex()` returns a sorted slice.
https://github.com/aquasecurity/fanal/blob/6b9b47f53e3f9ef4af7e4c750fc37d9c8a726827/secret/scanner.go#L104-L106
https://github.com/aquasecurity/fanal/blob/6b9b47f53e3f9ef4af7e4c750fc37d9c8a726827/secret/scanner.go#L116-L118

Let's say the first `regex.FindAllIndex()` returns `[][]int{{10, 13}, {15, 17}}` and the second `regex.FindAllIndex()` returns `[][]int{{2, 4}, {6, 7}}`. Then, `allowedLocations` will be `[][]int{{10, 13}, {15, 17}, {2, 5}, {6, 7}}`. And `StartPosition` is 3 and `EndPosition` is 4, for example. `10 > 4` will be true and the result is added into `results`. But it is actually within `{2, 5}` and should be allowed.

knqyf263 (commented):
The logic is a bit complicated now. Could you add unit tests?

secret/scanner.go
```diff
@@ -72,6 +134,38 @@ func (s Scanner) Scan(args ScanArgs) types.Secret {
...
+			results = append(results, result)
+		}
+	}
+	return results
+}
+
+func isResultAllowed(result RuleResult, allowedLocations [][]int) bool {
+	for _, allowedLocation := range allowedLocations {
+		if result.StartPosition >= allowedLocation[0] && result.EndPosition <= allowedLocation[1] || result.StartPosition <= allowedLocation[0] && result.EndPosition >= allowedLocation[1] {
+			return true
```
VaismanLior, line 163: I realized we need to check it both ways.
Gitleaks searches for the allowed patterns only inside the match, for example:
found secret - `password: "defaultPassword"`, and the allow list has the regex `default`, it will ignore it.

knqyf263:
Just in case you missed my comment, could you add unit tests in this PR?
https://github.com/aquasecurity/fanal/pull/443#pullrequestreview-931945811

VaismanLior:
> Just in case you missed my comment, could you add unit tests in this PR? [#443 (review)](https://github.com/aquasecurity/fanal/pull/443#pullrequestreview-931945811)

Sure, I'll add today

secret/scanner.go
```diff
@@ -11,24 +11,53 @@ import (
...
+	Title   string
+	Regexes []*regexp.Regexp
+}
+
+func NewScanner(rulePath string, rules []Rule, allowlist AllowList, excludeBlocks ExcludeBlocks) Scanner { // TODO: remove rules, allowlist, excludeBlocks when rulePath implemented
 	if rulePath != "" {
 		// TODO: Load custom rules here
 	}
+	if len(rules) == 0 {
+		rules = builtinRules
```
knqyf263, line 55: Should we disable built-in rules if a user passes even a single rule? I thought we wanted to merge user-specified rules into build-in rules.

secret/secret_scanner_test.go
```diff
@@ -0,0 +1,251 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := secret.NewScanner("", tt.fields.Rules, tt.fields.AllowList, tt.fields.ExcludeBlocks)
+			content, err := ioutil.ReadFile(tt.inputFilePath)
```
knqyf263, line 245: ioutil is deprecated.
```suggestion
			content, err := os.ReadFile(tt.inputFilePath)
```

secret/scanner.go
VaismanLior: It's only for the tests right now, I'll add reading from config YAML file and implement the merge in a different branch. 
I think we should merge the rules by the ruleID, so the user will be able to replace built-in rules if needed and add custom rules in a single config file.

secret/secret_scanner_test.go
```diff
@@ -0,0 +1,251 @@
...
+				FilePath: "testdata/secret.txt",
+				Findings: []types.SecretFinding{
+					{
+						RuleID:    basicRule.ID,
+						Type:      basicRule.Type,
+						Severity:  basicRule.Severity,
+						Title:     basicRule.Title,
+						StartLine: 2,
+						EndLine:   3,
+						Match:     "generic secret line *****",
```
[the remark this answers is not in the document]
VaismanLior: Yes, I've implemented it in the branch of the built-in rules.

secret/scanner.go
knqyf263: It sounds a good idea!

secret/secret_scanner_test.go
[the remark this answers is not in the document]
knqyf263: Cool. Noted.

secret/secret_scanner_test.go
VaismanLior: fixed

PR #452, opened 2022-04-04
chore(ci): enable gofmt linter

afdesk:
Gofmt checks whether code was gofmt-ed.


.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3.1.0
         with:
           version: v1.45
-          args : --verbose
+          args : --verbose --enable=gofmt
```
knqyf263, line 24: We should use `.golangci.yaml`.
https://github.com/aquasecurity/fanal/blob/main/.golangci.yml

afdesk: sure, missed it.

knqyf263 (approved):

PR #453, opened 2022-04-04
refactor(golang): use debug/buildinfo

knqyf263:
https://github.com/aquasecurity/go-dep-parser/pull/87


analyzer/language/golang/binary/binary_test.go
```diff
@@ -43,7 +43,6 @@ func Test_gobinaryLibraryAnalyzer_Analyze(t *testing.T) {
 		{
 			name:      "broken elf",
 			inputFile: "testdata/broken_elf",
-			wantErr:   "unexpected EOF",
```
afdesk, line 46: sorry for bother. 
`wantErr` is never used, right?

knqyf263: Thanks. You're right. I've removed it.

afdesk (approved):

PR #464, opened 2022-04-12
Secret builtin rules

VaismanLior:


secret/scanner.go
```diff
@@ -86,6 +93,21 @@ func (r *Rule) FindLocations(content []byte) []Location {
...
+	var indices [][]int
+	matchsLocs := r.Regex.FindAllSubmatchIndex(content, -1)
+	for _, matchLocs := range matchsLocs {
+		for i, name := range r.Regex.SubexpNames() {
+			if name == r.SecretGroupName {
+				startLocIndex := 2 * i
+				endLocIndex := startLocIndex + 1
+				indices = append(indices, []int{matchLocs[startLocIndex], matchLocs[endLocIndex]})
+			}
+		}
```
knqyf263, line 106: You may be able to use `SubexpIndex`

```suggestion
		startLocIndex := 2 * r.Regex. SubexpIndex(r.SecregGroupName)
		endLocIndex := startLocIndex + 1
		indices = append(indices, []int{matchLocs[startLocIndex], matchLocs[endLocIndex]})
```

VaismanLior: The only difference is that `SubexpNames` will support multiple `secret` groups in one regex.
While `SubexpIndex` will return only the first.
I guess it is nice to have but not a must.

knqyf263: Yes, `SubexpNames` looks like we support multiple groups, but we support a single group, right? If so, `SubexpIndex` is straightforward and easy to understand. Also, we can warn the wrong group name like

```
		if subexpIndex == -1 {
			log.Logger.Warnf("The specified group %s is not found in %s", r.SecregGroupName, r.ID)
			continue
		}
```

When we support multiple groups, we can switch to `SubexpNames`.

VaismanLior: We kind of support multiple groups, but with the same name.
For example for the regex -
`credentials:\s*\{\s*user:\s*['"](?P<secret>[0-9a-zA-Z\-_=]{8,10})['"]\s*password:\s*['"](?P<secret>[0-9a-zA-Z\-_=]{8,10})['"]\s*\}`

Over - 

`credentials: {	user: "username" password: "123456789" }`

It will return two results: 
 1. `credentials: { user: "*****" password: "123456789" }`
 2. `credentials: { user: "username" password: "*****" }`

knqyf263: Does the conflicting group names work?

VaismanLior: Yes, I ran a unit test with this example, and Go doesn't care.

knqyf263: Interesting. I didn't know it. You could capture both?

VaismanLior: Yep, it works.
I can add a test of that scenario as well.

knqyf263: I confirmed it. I'm not sure how useful it is in our case, but let's keep it then.
https://go.dev/play/p/xcWkE1zgpA0

Issue #468, opened 2022-04-13
Use /etc/os-release when present instead of /etc/alpine-release et al


kaniini:
The `ID` field in `/etc/os-release` is intended to be used for scanning tools to detect the OS.  Syft uses `/etc/os-release` and then falls back to other detection methods when not present.

PR #470, opened 2022-04-13
feat(analyzer): more general support for os-release

knqyf263:
Close https://github.com/aquasecurity/fanal/issues/468


walker/tar.go
```diff
@@ -64,7 +65,15 @@ func (w LayerTar) Walk(layer io.Reader, analyzeFn WalkFunc) ([]string, []string,
...
 			}
-		case tar.TypeSymlink, tar.TypeLink, tar.TypeReg:
+		case tar.TypeSymlink:
+			if filePath == "etc/os-release" {
+				osReleaseIsSymLink = true
+				continue
+			}
+			if w.shouldSkipFile(filePath) {
+				continue
+			}
```
knqyf263, line 75: I don't think we need this process. We can just add `usr/lib/os-release` to `requiredFiles` in osReleaseAnalyzer.

DmitriyLewen: Done

PR #510, opened 2022-05-03
refactor(secret): allow list match with secret subgroups

VaismanLior:
When using the secret-group-name feature, the match we check the allow-list-rules on is the secret-group itself.
This behavior is unexpected and causes more false positives.
Changed to match the whole regex match for allow-rules, not only the secret-group.


knqyf263 (commented):
Can we fix the existing test so that we can make sure this change is correct? Also, could you add the issue detail in the PR description?

knqyf263 (approved):

PR #511, opened 2022-05-03
Pom modules

ShiraCohen33:
fix bug https://github.com/aquasecurity/trivy/issues/2083


knqyf263 (changes requested):
Could you add a test so we can make sure this change fixes an issue?

knqyf263 (approved):

PR #513, opened 2022-05-04
fix(config): skip json config if type is not detected

DmitriyLewen:
## Description
Fanal saves all json file from system as config files.
If there are a lot of json files in the system, it can cause a lot of memory usage.
Added skip json files without config type.


analyzer/config/json/json.go
```diff
@@ -30,21 +31,43 @@ func NewConfigAnalyzer(filePattern *regexp.Regexp) ConfigAnalyzer {
...
 		return nil, xerrors.Errorf("unable to decode JSON (%s): %w", input.FilePath, err)
 	}
 
-	return &analyzer.AnalysisResult{
-		Configs: []types.Config{
-			{
-				Type:     types.JSON,
-				FilePath: input.FilePath,
-				Content:  parsed,
+	if configs, ok := parsed.([]interface{}); ok { // json input is array
```
knqyf263, line 42: JSON doesn't need this sort of process as it is supported by YAML only.

DmitriyLewen: I've updated logic with changes from man branch. Can you take a look?

analyzer/config/json/json.go
```diff
@@ -36,18 +38,23 @@ func (a ConfigAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput)
...
-	return &analyzer.AnalysisResult{
-		Files: map[types.HandlerType][]types.File{
-			// It will be passed to misconfig post handler
-			types.MisconfPostHandler: {
-				{
-					Type:    types.JSON,
-					Path:    input.FilePath,
-					Content: b,
+	for _, supportedType := range supportedTypes { // system can contain many json files. Ignore unsupported types.
+		if detection.IsType(input.FilePath, bytes.NewReader(b), supportedType) {
```
knqyf263, line 42: Looks good, but we can consider how to avoid calling `IsType` twice.
https://github.com/aquasecurity/fanal/blob/03d2949f6ff52825d8fb55bf2bc3795330a0f84b/handler/misconf/misconf.go#L211-L213

I think we can put the detailed type to `Type` instead of `types.JSON`, then skip calling `IsType` here if it is the detailed type.
https://github.com/aquasecurity/fanal/blob/03d2949f6ff52825d8fb55bf2bc3795330a0f84b/handler/misconf/misconf.go#L211-L213

knqyf263 (commented):
@liamg Could you also review this PR? You are more familiar now 😄

liamg:
I think it might be a nice approach to refactor `types.File` to instead hold a function which provider a reader to a file, instead of holding the raw `[]byte` slice itself. That way we only read files into memory when we need to process them. WDYT?

knqyf263:
I did it at the beginning because that is why I asked you to add LazyMemoryFS. After some refactoring, I switched to `[]byte` for some reason. But it is worth trying it again. @liamg Could you take it over?
@DmitriyLewen Let me close this PR. Thanks.

PR #514, opened 2022-05-05
fix(secrets): skip aws secrets of greater length

DmitriyLewen:
## Description

Changed regex for aws-secret-access-key:
- added prefix check: currently we get all secrets with `key` prefix(like `publicKey=*****`).
- added secret length check. We should get secrets with only 40 characters after `=`(like `aws_key=*****asdWef`).

## Related Issues
- aquasecurity/trivy/issues/2060


secret/builtin-rules.go
```diff
@@ -78,7 +88,7 @@ var builtinRules = []Rule{
 		Category:        CategoryAWS,
 		Severity:        "CRITICAL",
 		Title:           "AWS Secret Access Key",
-		Regex:           MustCompile(`(?i)["']?(aws)?_?(secret)?_?(access)?_?key["']?\s*(:|=>|=)\s*(?P<secret>["']?[A-Za-z0-9\/\+=]{40})["']?`),
+		Regex:           MustCompile(`(?i)` + startSecret + quote + aws + `(secret)?_?(access)?_?key` + quote + connect + quote + `(?P<secret>[A-Za-z0-9\/\+=]{40})` + quote + endSecret),
```
knqyf263, line 91: Can we use fmt.Sprintf?

knqyf263 (commented):
We should add tests for legitimate AWS key and invalid values like Base64.

secret/builtin-rules.go
DmitriyLewen: Done

knqyf263 (approved):

PR #525, opened 2022-05-12
fix(secrets): added leading slash for file paths extracted from image

DmitriyLewen:
## Description 
Added leading `/` for file paths extracted from `image` when analyzing secrets.
The working directory for container image is always `/`. For `path` in secret configuration file, it is better to use an absolute path.

## Related issues
- aquasecurity/trivy/issues/2060

### DIFF
--- a/analyzer/secret/secret.go
+++ b/analyzer/secret/secret.go
@@ -2,6 +2,7 @@ package secret
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"math"
 	"os"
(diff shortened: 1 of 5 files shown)


analyzer/secret/secret.go
```diff
@@ -75,8 +78,13 @@ func (a SecretAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput)
 		return nil, xerrors.Errorf("read error %s: %w", input.FilePath, err)
 	}
 
+	filePath := input.FilePath
```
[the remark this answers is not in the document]
DmitriyLewen: Looks good! I did this

analyzer/secret/testdata/image-config.yaml
```diff
@@ -0,0 +1,9 @@
+rules:
```
knqyf263, line 1: I thought we would add `path` here so that we can make sure our allowrule works with image scanning.

DmitriyLewen: Done

knqyf263 (approved):

PR #534, opened 2022-05-20
feat: adding helm support

owenrumney:
- feat: adding helm support
- adding tests for helm analyzer
- add test for non helm tarball
- adding in-memory filesystem for helm
- handle multiple charts at a time
- chore(deps): bump defsec version


analyzer/config/helm/helm.go
```diff
@@ -0,0 +1,132 @@
...
+func (a ConfigAnalyzer) Required(filePath string, _ os.FileInfo, readerOpener analyzer.Opener) bool {
+	if a.filePattern != nil && a.filePattern.MatchString(filePath) {
+		return true
+	}
+
+	if isArchive(filePath) {
+		reader, err := readerOpener()
+		if err != nil {
+			return false
+		}
```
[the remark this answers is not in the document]
owenrumney: done

liamg (approved):

analyzer/config/helm/helm.go
```diff
@@ -0,0 +1,133 @@
...
+		return true
+	}
+
+	if isArchive(filePath) {
+		reader, err := readerOpener()
+		defer func() { _ = reader.Close() }()
+		if err != nil {
+			return false
+		}
+		return isHelmChart(filePath, reader)
```
knqyf263, line 62: Can we do this detection in `Analyze()`? Then, we don't have to pass `opener` to `Required`.

owenrumney: Yeah, I appreciate not ideal to pass in the opener however this was so we cool look inside the tarballs and see if they were helm charts. I'm trying to avoid adding any found `tar`, `tgz`, or `tar.gz` that is found being added into the scanned file system.

I can look to move it into the Anaylze but I'm concerned about adding random very large tarballs that aren't needed.

owenrumney: It might be okay if I change https://github.com/aquasecurity/fanal/blob/main/analyzer/analyzer.go#L311-L322 to handle `nil` being returned and not try an merge. then the `IsHelmChart` logic can move. 

If this is preferable, I'll do that

knqyf263: `opener` returns *os.File.
https://github.com/aquasecurity/fanal/blob/ca3bf967cb0d97a8a9939ffd0a702c2fa5fa904b/walker/fs.go#L66-L70

And `isHelmChart` processes tar.gz streamingly with gzip.NewReader and tar.NewReader. I believe we don't load the tarball into memory unless we detect it as Helm chart.
https://github.com/aquasecurity/fanal/blob/fbb8cf9d9e598d2ee31fe513efe38282ce8fe5ac/analyzer/config/helm/helm.go#L90-L116

 If your 10GB photo tarball includes `Chart.yaml`, it will be problematic.
https://github.com/aquasecurity/fanal/blob/fbb8cf9d9e598d2ee31fe513efe38282ce8fe5ac/analyzer/config/helm/helm.go#L111-L113

knqyf263: But we can skip too large tarball anyway because it is probably not Helm chart.

owenrumney: added an arbitrary size gate of 200mb and moved the IsHelm call into the Analyze function. 

A lot less intrusive change now

analyzer/config/helm/helm.go
```diff
@@ -0,0 +1,141 @@
...
+func isArchive(path string) bool {
+	if strings.HasSuffix(path, ".tar") ||
+		strings.HasSuffix(path, ".tgz") ||
+		strings.HasSuffix(path, ".tar.gz") {
+		return true
+	}
+	return false
+}
+
+func isZip(path string) bool {
```
knqyf263, line 135:
```suggestion
func isGzip(path string) bool {
```

analyzer/config/helm/helm.go
```diff
@@ -0,0 +1,141 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func isArchive(path string) bool {
+	if strings.HasSuffix(path, ".tar") ||
+		strings.HasSuffix(path, ".tgz") ||
+		strings.HasSuffix(path, ".tar.gz") {
```
knqyf263, line 129: nit
```suggestion
		isGzip(path)
```

knqyf263 (approved):
Looks good to me other than the place of size check. Approved.

analyzer/config/helm/helm.go
```diff
@@ -0,0 +1,141 @@
...
+	return ConfigAnalyzer{
+		filePattern: filePattern,
+	}
+}
+
+func (a ConfigAnalyzer) Analyze(_ context.Context, input analyzer.AnalysisInput) (*analyzer.AnalysisResult, error) {
+	if input.Info.Size() > maxTarSize {
+		// tarball is too big to be Helm chart - move on
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
owenrumney: yup, that's a good point. fixed.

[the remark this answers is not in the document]
knqyf263: Thanks! You can merge it now.

knqyf263:
Ref. https://github.com/aquasecurity/trivy/issues/1757

PR #535, opened 2022-05-27
refactor(deps): move dependencies to package

AndreyLevchenko:
moving dependencies into packages

### DIFF
--- a/types/artifact.go
+++ b/types/artifact.go
@@ -43,7 +43,8 @@ type Package struct {
 	Layer    Layer  `json:",omitempty"`
 
 	// Each package metadata have the file path, while the package from lock files does not have.
-	FilePath string `json:",omitempty"`
+	FilePath     string                  `json:",omitempty"`
+	Dependencies []godeptypes.Dependency `json:"Dependencies,omitempty"`
 }
 
 // BuildInfo represents information under /root/buildinfo in RHEL
--- a/types/artifact.go
+++ b/types/artifact.go
@@ -3,7 +3,6 @@ package types
 import (
 	"time"
 
-	godeptypes "github.com/aquasecurity/go-dep-parser/pkg/types"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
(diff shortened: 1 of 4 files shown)


types/artifact.go
```diff
@@ -43,7 +43,8 @@ type Package struct {
 	Layer    Layer  `json:",omitempty"`
 
 	// Each package metadata have the file path, while the package from lock files does not have.
-	FilePath string `json:",omitempty"`
+	FilePath     string                  `json:",omitempty"`
+	Dependencies []godeptypes.Dependency `json:"Dependencies,omitempty"`
```
knqyf263, line 47: I think it is `DependsOn []string`, no?

AndreyLevchenko: You are correct it's more concise. But I can't build a dependency graph from it.

knqyf263: Interesting. Why not?

AndreyLevchenko: `PkgID` and list of dependent ids are stored. So it looks like everything is direct dependency.

knqyf263: Currently, we have only direct dependencies in `Dependency`, right?
https://github.com/aquasecurity/go-dep-parser/blob/d316f5cc2cff93431f2ff8986ffbaf9a321da512/pkg/types/types.go#L13-L16

```
[]Dependency {
    {
        ID: "app1@1.0.0",
        DependsOn: {
            "app2@2.0.0",
            "app3@3.0.0",
        },
    },
    {
        ID: "app2@2.0.0",
        DependsOn: {
            "app4@4.0.0",
        },
    },
}
```

We just change it to

```
[]Package {
    {
        ID: "app1@1.0.0",
        Name: "app1",
        Version: "1.0.0",
        License: "MIT",
        DependsOn: {
            "app2@2.0.0",
            "app3@3.0.0",
        },
    },
    {
        ID: "app2@2.0.0",
        Name: "app2",
        Version: "2.0.0",
        License: "MIT",
        DependsOn: {
            "app4@4.0.0",
        },
    },
}
```

The information is the same, no?

knqyf263: `app4@4.0.0` is an indirect dependency of `app1@1.0.0`. We can know it from `DependsOn`.

knqyf263:
@AndreyLevchenko We have to remove `Dependencies`.
https://github.com/aquasecurity/fanal/blob/9a677d613f022abf9dbdb00681fbe810563c299a/types/artifact.go#L82-L83

knqyf263 (approved):

Issue #549, opened 2022-06-01
feat(dotnet): Add support for parsing .deps.json files for .Net core


astevenson-microsoft:
.Net core generates a [appname].deps.json alongside the executable which contains the dependencies of that executable. Parsing it would provide information on dependencies for deployed .Net core executables. Support for parsing .deps.json was added to go-dep-parser in https://github.com/aquasecurity/go-dep-parser/pull/103.

PR #550, opened 2022-06-01
feat(dotnet): Add support for .Net core .deps.json files (#549)

astevenson-microsoft:
feat(dotnet): Add support for .Net core .deps.json files.

Fixes https://github.com/aquasecurity/fanal/issues/549


knqyf263 (commented):
The test was failing not because of your change, so I've fixed it. Could you add a test?

Issue #562, opened 2022-06-13
support rbac scanner and type


chen-keinan:
it is require to add support for rbac scanner and rbac type

PR #563, opened 2022-06-13
feat: support rbac scanner and type

chen-keinan:
Signed-off-by: chenk <hen.keinan@gmail.com>
support rbac scanner and type - solve #562


knqyf263:
Looks like defsec related. I requested a review from Liam instead.

